PR #22, opened 2021-07-13
proxmox-clone: choose template located on target node if possible

mizhka:
In case of multiple templates with same name on multiple hosts, proxmox-clone chooses first template
returned by Proxmox. This commit forces it to choose template located on
target node if possible. It's helpful if templates use local storage and it's
impossible to clone templates between nodes, i.e. to avoid errors like:
"500 can't clone VM to node 'host2' (VM uses local storage)"
or
"500 storage 'xyz' is not available on node 'host1'"

### DIFF
--- a/builder/proxmox/clone/builder.go
+++ b/builder/proxmox/clone/builder.go
@@ -58,10 +58,19 @@ func (*cloneVMCreator) Create(vmRef *proxmoxapi.VmRef, config proxmoxapi.ConfigQ
 	config.FullClone = &fullClone
 	config.CIuser = comm.SSHUsername
 	config.Sshkeys = string(comm.SSHPublicKey)
-	sourceVmr, err := client.GetVmRefByName(c.CloneVM)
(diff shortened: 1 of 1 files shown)


SwampDragons:
cc @carlpett

mizhka:
Hi, @carlpett , @SwampDragons ,

It is possible to consider this PR? 
The regular errors "Error creating VM: 500 can't clone VM to node " makes team crying. 
It will be nice to have feedback on it: negative or positive. Thank you in advance!

carlpett (commented):
Hi @mizhka,
Apologies for the lack of response. Due to things happening, I've had to take a longer absence than planned, but I will hopefully be available from now on.
The changes look reasonable, so nice work! I do not have a multi-node setup to test it on, unfortunately, but it should not be able to fail from what I can tell.
Could you just rebase it to get rid of the merge conflict?

mizhka:
@carlpett , sorry for delay.
Commit is updated to latest main branch state.

PR #26, opened 2021-08-03
Fix nav_title for clone builder documentation

AaronJackson:
On the left hand menu, both proxmox-clone and proxmox-iso appeared as "ISO"

image


karliemeads (approved):

azr (approved):
Thanks ! LGTM

PR #27, opened 2021-08-06
Allow user to specify task_timeout

### LINKED ISSUE #20: Client default timeout needs adjustment for longer clones
I believe the default timeout period (30 sec) is being used to wait for completion of a full clone.  When clones take longer than this, I'm seeing timeout errors, I'm guessing this is because my proxmox server is slower or the disks being cloned are relatively large.  When not performing a full_clone, the timeout is not reached and the plugin behaves normally.  Clones in my case take 90 seconds.

Here is where the default is set:
https://github.com/hashicorp/packer-plugin-proxmox/blob/de55cac730fd9626080569bde50c89ae871b2031/builder/proxmox/common/client.go#L11

Which eventually gets passed here where the error occurs:
https://github.com/Telmate/proxmox-api-go/blob/51bd4ae21d9a8ecdd7f040c48c240756616c7d08/proxmox/client.go#L577

Ideally I could imagine this being broken out as an explicit configuration parameter.  A more expedient fix is to just increase the default timeout.

Packer v1.7.2
packer-plugin-proxmox 0.0.2

Relevant log fragment:

```
==> proxmox-clone.autogenerated_1: Creating VM
==> proxmox-clone.autogenerated_1: No VM ID given, getting next free from Proxmox
2021/06/06 10:32:44 packer-builder-proxmox-clone plugin: 500 can't lock file '/var/lock/qemu-server/lock-105.conf' - got timeout
==> proxmox-clone.autogenerated_1: Error creating VM: 500 can't lock file '/var/lock/qemu-server/lock-105.conf' - got timeout
2021/06/06 10:32:44 [INFO] (telemetry) ending 
==> Wait completed after 40 seconds 622 milliseconds
2021/06/06 10:32:44 machine readable: error-count []string{"1"}
==> Some builds didn't complete successfully and had errors:
2021/06/06 10:32:44 machine readable: proxmox-clone.autogenerated_1,error []string{"Error creating VM: 500 can't lock file '/var/lock/qemu-server/lock-105.conf' - got timeout"}
==> Builds finished but no artifacts were created.
2021/06/06 10:32:44 [INFO] (telemetry) Finalizing.
Build 'proxmox-clone.autogenerated_1' errored after 40 seconds 622 milliseconds: Error creating VM: 500 can't lock file '/var/lock/qemu-server/lock-105.conf' - got timeout

==> Wait completed after 40 seconds 622 milliseconds

==> Some builds didn't complete successfully and had errors:
--> proxmox-clone.autogenerated_1: Error creating VM: 500 can't lock file '/var/lock/qemu-server/lock-105.conf' - got timeout

==> Builds finished but no artifacts were created.
```

I'm having trouble explaining why the error occurs consistently after 40 seconds and not 30 seconds.  It is possible I'm mixing up the timeout at fault here.

AaronJackson:
Closes #20 

The default timeout of 30 seconds is too slow for many proxmox setups. This changes the default to 60 seconds, as well as introduces a new builder argument called `task_timeout`, specified as a duration string.

### DIFF
--- a/builder/proxmox/clone/config.hcl2spec.go
+++ b/builder/proxmox/clone/config.hcl2spec.go
@@ -85,6 +85,7 @@ type FlatConfig struct {
 	Token                     *string                            `mapstructure:"token" cty:"token" hcl:"token"`
 	Node                      *string                            `mapstructure:"node" cty:"node" hcl:"node"`
 	Pool                      *string                            `mapstructure:"pool" cty:"pool" hcl:"pool"`
+	TaskTimeout               *string                            `mapstructure:"task_timeout" cty:"task_timeout" hcl:"task_timeout"`
 	VMName                    *string                            `mapstructure:"vm_name" cty:"vm_name" hcl:"vm_name"`
 	VMID                      *int                               `mapstructure:"vm_id" cty:"vm_id" hcl:"vm_id"`
 	Boot                      *string                            `mapstructure:"boot" cty:"boot" hcl:"boot"`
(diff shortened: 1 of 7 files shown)


azr (approved):
Nice one, LGTM, thanks !

PR #33, opened 2021-09-03
add cd_files support to additional_iso_files config

quasmon:
Adds support for `cd_files` from the packer-plugin-sdk for #8. Currently only added it to the `additional_iso_files` config


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/hashicorp/packer-plugin-proxmox
 go 1.17
 
 require (
-	github.com/Telmate/proxmox-api-go v0.0.0-20210402150908-b9cc6607c070
+	github.com/Telmate/proxmox-api-go v0.0.0-20210825163308-5e4c0d698a78
```
quasmon, line 6: Bumping dependency to get access to DeleteVolume to cleanup generate ISO files. This also fixes virtual machines not correctly being deleted (noticed on my Proxmox instance v7.0-9)

main.go
```diff
@@ -14,8 +14,8 @@ import (
 func main() {
 	pps := plugin.NewSet()
 	pps.RegisterBuilder(plugin.DEFAULT_NAME, new(proxmoxiso.Builder))
-	pps.RegisterBuilder("proxmox-iso", new(proxmoxiso.Builder))
-	pps.RegisterBuilder("proxmox-clone", new(proxmoxclone.Builder))
+	pps.RegisterBuilder("iso", new(proxmoxiso.Builder))
+	pps.RegisterBuilder("clone", new(proxmoxclone.Builder))
 	pps.SetVersion(version.PluginVersion)
```
azr, line 19: Good catch, this might be a breaking change though!

builder/proxmox/common/step_upload_additional_iso.go
```diff
@@ -0,0 +1,90 @@
...
+
+	p := state.Get(s.ISO.DownloadPathKey).(string)
+	if p == "" {
+		err := fmt.Errorf("path to downloaded ISO was empty")
+		state.Put("error", err)
+		ui.Error(err.Error())
+		return multistep.ActionHalt
+	}
+
+	isoPath, _ := filepath.EvalSymlinks(p)
```
azr, line 49: Could be a good idea to not ignore that error to avoid weird/unexpected bugs

azr (approved):
Super nice PR, loved the docs change. I'll let @carlpett have the final word.

builder/proxmox/common/step_upload_additional_isos.go
```diff
@@ -1,68 +0,0 @@
...
-
-// stepUploadAdditionalISOs uploads all additional ISO files that are mountet
-// to the VM
-type stepUploadAdditionalISOs struct{}
-
-type uploader interface {
-	Upload(node string, storage string, contentType string, filename string, file io.Reader) error
-}
-
-var _ uploader = &proxmox.Client{}
```
carlpett, line 23: This is a structural helper for making sure the step is testable. Apparently we didn't add any tests for this step (but in the original proxmox-iso builder there are tests for the "single ISO upload" step) when it was added, though.
It'd be nice to have a simple happy case test at least for the new step, or even just keeping the testability.

carlpett (commented):
Hi @quasmon,
Thanks for working on this! Functionally looks good to me, and as mentioned very nice work on the docs :clap:
If you'd be willing to look at a test as I note in the comment below, that'd be fantastic. There's a fair number of cases.

Still, overall very nice, and I think restructuring it to have the step deal with a single file and running it multiple times was a very good idea!

Tristan971:
Hi, this PR made me really happy, and looks like a really clean approach, thanks a ton and very much looking forward to see it upstreamed.

I would however wonder whether to keep the `cd_content` part of the docs considering this MR doesn't actually add support for it... 😅

quasmon:
> I would however wonder whether to keep the `cd_content` part of the docs considering this MR doesn't actually add support for it... sweat_smile

Thanks @Tristan971, and excellent find! Completely missed `cd_content`. I have now included that functionality

builder/proxmox/common/step_upload_additional_isos.go
quasmon: Thanks for spotting that one, wasn't too sure about the testing when I initially committed.

I've now updated this PR with some test coverage which follow the "single ISO upload" step.

carlpett:
Nice work on the tests! I think this all looks good to go, apart from the worry @azr mentioned about renaming the builders.
Can we take that change in a separate PR?

quasmon:
> Nice work on the tests! I think this all looks good to go, apart from the worry @azr mentioned about renaming the builders. Can we take that change in a separate PR?

All done, I have removed the rename

quasmon:
Had one last bug fix for the parser config logic when incorrectly specifying multiple sources like `cd_files` and `iso_url`, have also provided appropriate tests as well.

carlpett (approved):
Great work, thanks for quick fixes! :cake:

azr (approved):
Super nice, thanks for adressing the changes !

PR #57, opened 2022-01-02
Clarify requirements for token authentication

ProfessorSalty:
The text describing how to set the username when using token authentication is confusing as it makes no mention of the token id, which is what needs to be appended to the username.


carlpett (approved):
Thanks!

PR #61, opened 2022-01-18
Use VM name when template name not provided

sudomateo:
The Proxmox Packer plugin documentation states this:

> template_name (string) - Name of the template. Defaults to the generated name used during creation.

However, during testing, when the `template_name` option is not provided the following error occurs:

```
==> centos.proxmox-iso.centos: Error updating template: 500 no options specified
```

These changes default to using the `VMName` from the configuration, layering in the user provided `template_name` when it exists.


nywilken (approved):
LGTM. Thanks for the fix.

PR #72, opened 2022-03-07
[Issue 71] Adds convert_to_template option to proxmox builders

### LINKED ISSUE #71: Add "convert_to_template"  option to iso and clone builders
Please search the existing issues for relevant feature requests, and use the
reaction feature
(https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)
to add upvotes to pre-existing requests.

#### Community Note

#### Description
This feature will add a "convert_to_template" option to the proxmox-iso and proxmox-clone builders. The option will instruct this plugin to keep the machine image as a VM or convert it to a template. 

#### Use Case(s)

Gives the user more control over the created machine image. 

#### Potential configuration

```
convert_to_template = true|false
```

#### Potential References

The packer vsphere-iso and vsphere-clone builders have similar functionality 

https://www.packer.io/plugins/builders/vsphere/vsphere-iso#convert_to_template

LethalServant:
Adds `convert_to_template` bool option to proxmox-iso and proxmox-clone builders configuration. 
This gives the user more control over the machine image, whether it stays a VM or is converted to a VM Template. 
The `convert_to_template` option is optional, and the default value is `false`, resulting in the VM not being converted to a template.
In practice the option is only need if the user wants the machine image to become a VM Template.

Tests:
Given working proxmox-iso and proxmox-clone packer file configurations, the following values were changed and run for each builder:

```
convert_to_template      = true  // result: vm is a template
convert_to_template      = false // result: vm is not a template 
remove convert_to_template option, testing default value // result: vm is not a template
```

Closes #71

### DIFF
--- a/builder/proxmox/common/builder.go
+++ b/builder/proxmox/common/builder.go
@@ -67,9 +67,7 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook,
 	})
 	coreSteps = append(coreSteps, &stepStopVM{})
 
-	if b.config.ConvertToTemplate {
-		coreSteps = append(coreSteps, &stepConvertToTemplate{})
-	}
+	coreSteps = append(coreSteps, &stepConvertToTemplate{})
 
 	coreSteps = append(coreSteps, &stepFinalizeVMConfig{})
 	coreSteps = append(coreSteps, &stepSuccess{})
--- a/builder/proxmox/common/builder.go
+++ b/builder/proxmox/common/builder.go
@@ -108,7 +108,7 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook,
 	artifact := &Artifact{
 		builderID:     b.id,
 		vmID:          vmRef.VmId(),
-		isTemplate:    b.config.ConvertToTemplate,
+		isTemplate:    !b.config.SkipConvertToTemplate.True(),
 		proxmoxClient: b.proxmoxClient,
 		StateData:     map[string]interface{}{"generated_data": state.Get("generated_data")},
 	}
--- a/builder/proxmox/common/builder.go
+++ b/builder/proxmox/common/builder.go
@@ -47,29 +47,33 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook,
 
 	comm := &b.config.Comm
 
-	// Build the steps
-	coreSteps := []multistep.Step{
-		&stepStartVM{
-			vmCreator: b.vmCreator,
-		},
-		commonsteps.HTTPServerFromHTTPConfig(&b.config.HTTPConfig),
-		&stepTypeBootCommand{
-			BootConfig: b.config.BootConfig,
-			Ctx:        b.config.Ctx,
-		},
-		&communicator.StepConnect{
-			Config:    comm,
-			Host:      commHost((*comm).Host()),
-			SSHConfig: (*comm).SSHConfigFunc(),
-		},
-		&commonsteps.StepProvision{},
-		&commonsteps.StepCleanupTempKeys{
-			Comm: &b.config.Comm,
-		},
-		&stepConvertToTemplate{},
-		&stepFinalizeTemplateConfig{},
-		&stepSuccess{},
+	// Build the steps (Order matters)
+	coreSteps := []multistep.Step{}
+
+	coreSteps = append(coreSteps, &stepStartVM{vmCreator: b.vmCreator})
+	coreSteps = append(coreSteps, commonsteps.HTTPServerFromHTTPConfig(&b.config.HTTPConfig))
+	coreSteps = append(coreSteps, &stepTypeBootCommand{
+		BootConfig: b.config.BootConfig,
+		Ctx:        b.config.Ctx,
+	})
+	coreSteps = append(coreSteps, &communicator.StepConnect{
+		Config:    comm,
+		Host:      commHost((*comm).Host()),
+		SSHConfig: (*comm).SSHConfigFunc(),
+	})
+	coreSteps = append(coreSteps, &commonsteps.StepProvision{})
+	coreSteps = append(coreSteps, &commonsteps.StepCleanupTempKeys{
+		Comm: &b.config.Comm,
+	})
+	coreSteps = append(coreSteps, &stepStopVM{})
+
+	if b.config.ConvertToTemplate {
+		coreSteps = append(coreSteps, &stepConvertToTemplate{})
 	}
+
+	coreSteps = append(coreSteps, &stepFinalizeVMConfig{})
+	coreSteps = append(coreSteps, &stepSuccess{})
+
 	preSteps := b.preSteps
 	for idx := range b.config.AdditionalISOFiles {
 		preSteps = append(preSteps, &commonsteps.StepDownload{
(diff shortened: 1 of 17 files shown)


carlpett:
Hi @LethalServant,
So this would change the default behaviour of the builder quite a bit, is there a compelling reason for this?

sylviamoss:
@LethalServant do you have any update on @carlpett's question?

adrian-hoasted:
It should keep the default behaviour of converting into a template.
and if the option is set, it should change the default behaviour to leave the VM.

ie.
The convert_to_template option is optional, and the default value is **TRUE**, resulting in the VM being converted to a template. If the value is set to FALSE, no template is created. This would prevent a breaking change.

builder/proxmox/common/config.go
```diff
@@ -64,6 +64,7 @@ type Config struct {
 
 	TemplateName        string `mapstructure:"template_name"`
 	TemplateDescription string `mapstructure:"template_description"`
+	ConvertToTemplate   bool   `mapstructure:"convert_to_template"`
```
sylviamoss, line 67: The default of this configuration will always be false which will indeed change the default behavior of the builder. Also, in the `prepare` method you're explicitly setting it to false as default. 

If the intention is to not change the default behavior, the default should be `true`. In this case, you can: 
- Use [Trilian](https://github.com/hashicorp/packer-plugin-sdk/blob/main/template/config/custom_types.go#L8), a custom type available in the SDK for this type of configuration
- Rename it to `skip_convert_to_template` that will skip converting to template when set to true. The default will be automatically false.

builder/proxmox/common/builder.go
```diff
@@ -47,29 +47,33 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook,
...
+		SSHConfig: (*comm).SSHConfigFunc(),
+	})
+	coreSteps = append(coreSteps, &commonsteps.StepProvision{})
+	coreSteps = append(coreSteps, &commonsteps.StepCleanupTempKeys{
+		Comm: &b.config.Comm,
+	})
+	coreSteps = append(coreSteps, &stepStopVM{})
+
+	if b.config.ConvertToTemplate {
+		coreSteps = append(coreSteps, &stepConvertToTemplate{})
```
sylviamoss, line 71: I like better when we pass the configuration to the step and within the step is determined if it should be skipped or not based on the configuration. It makes it easier to log "Skipping ..." and keeps this step slice clean.

sylviamoss (changes requested):
I see this is indeed changing the default behavior and I figure that's not your intention as you mentioned in our previous comment.

builder/proxmox/common/config.go
LethalServant: @sylviamoss Thanks for the feedback, this should be resolved now.

zakx:
It's been two months, any update on this? Would love to have this functionality.

nywilken:
> It's been two months, any update on this? Would love to have this functionality.

Hi @LethalServant thanks for working to push this forward. We haven't had a chance to review this since you last updated. But I'm adding it to our planning board so we can give it some cycles next week. Quickly looking at the approach I don't know if `skip_convert_to_template` is the best way to avoid changing the default behavior. 

I will be releasing a new version of the plugin today v1.1.0 to address #119. After that change we can work to get this change in. Thanks again for pushing this change forward and for sticking with us.

builder/proxmox/iso/config_test.go
```diff
@@ -109,6 +110,9 @@ func TestBasicExampleFromDocsIsValid(t *testing.T) {
 	if b.config.CloudInit != false {
 		t.Errorf("Expected CloudInit to be false, got %t", b.config.CloudInit)
 	}
+	if b.config.SkipConvertToTemplate.False() {
+		t.Errorf("Expected ConvertToTemplate to be true, got %t", b.config.SkipConvertToTemplate.True())
```
lbajolet-hashicorp, line 114: This should be rephrased, also I suspect the error was introduced in the first draft, so the value does not match what the condition highlights.

builder/proxmox/iso/config_test.go
```diff
@@ -42,11 +42,12 @@ func TestBasicExampleFromDocsIsValid(t *testing.T) {
...
       "ssh_password": "packer",
-
+	
       "unmount_iso": true,
       "template_name": "fedora-29",
-      "template_description": "Fedora 29-1.2, generated on {{ isotime \"2006-01-02T15:04:05Z\" }}"
-    }
+      "template_description": "Fedora 29-1.2, generated on {{ isotime \"2006-01-02T15:04:05Z\" }}",
+	  "skip_convert_to_template": true
+	}
```
lbajolet-hashicorp, line 50: The indentation is not consistent with the rest of the template here, the verbatim template being only indented with spaces.

builder/proxmox/common/step_stop_vm.go
```diff
@@ -0,0 +1,39 @@
...
+
+type stepStopVM struct{}
+
+type vmTerminator interface {
+	ShutdownVm(*proxmox.VmRef) (string, error)
+}
+
+var _ templateConverter = &proxmox.Client{}
+
+func (s *stepStopVM) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
```
lbajolet-hashicorp, line 22: I wonder why this becomes a step on its own here, it seems that the `step_convert_to_template` does this already, why not move the check after the VM is shutdown so we don't have to introduce another step?
For the record, I don't hate this change by itself, it's just that this brings more changes to this PR which makes the diff larger and harder to understand.

builder/proxmox/common/step_stop_vm.go
```diff
@@ -0,0 +1,39 @@
...
+	ui := state.Get("ui").(packersdk.Ui)
+	client := state.Get("proxmoxClient").(vmTerminator)
+	vmRef := state.Get("vmRef").(*proxmox.VmRef)
+
+	ui.Say("Stopping VM")
+	_, err := client.ShutdownVm(vmRef)
+	if err != nil {
+		err := fmt.Errorf("Error converting VM to template, could not stop: %s", err)
```
lbajolet-hashicorp, line 30: If we keep this step, the error message should be reworked to specify that this is not converting the VM to a template that failed, but the "stop VM" step.

lbajolet-hashicorp (changes requested):
Hi @LethalServant,

First off, apologies for letting this PR open without activity for so long, let's get the ball rolling on this one and get this feature in!

I left a couple of comments on your code, I think the `Trilean` is not necessary in the current state as you inverted the condition, if it can safely default to `false`, we can keep it as a `bool`, which simplifies the interactions with it later on.

Also the code is out of sync since it was untouched for a while, so you will unfortunately need to rebase on top of the current main if we want to merge this later.

Please let me know if you're interested in continuing this work, and if you need help also feel free to reach out, I'll surely be able to help.

Thanks!

lbajolet-hashicorp:
Note: removed the `bump minor` label since in the current state, this does not change how the plugin works, so this can be a patch release.

mfisher87:
The author of this PR hasn't had any GitHub activity in a year or two. Is there a path forward to merge this?

tacoman:
Would love to see this merged in for our current workflow.

lbajolet-hashicorp:
Hi @AnthonyTippy @tacoman @mfisher87,

This looks pretty stale still, haven't heard of feedback from @LethalServant yet (to be fair to them, we waited a really long time for reviewing the code so there's definitely a lot of blame on us).

If someone wants to pickup that PR and add contributions on top to address the comments (that will likely need to be in a separate PR to supersede this one) I posted in my review, that would be very appreciated for sure, please let us know and we'll make sure to review the changes with minimal delay this time around.

LethalServant:
Hey @lbajolet-hashicorp 

I started working on this feature back around v1.0.4 so I think the code base has changed a lot and it would take some effort to get this back in mergeable state.  Not sure when I'll be able to get to this, if someone else is in a better position to add this feature either by picking this up or starting from scratch, go for it. Cheers!

lbajolet-hashicorp:
Since this one was left aside by @LethalServant's admission, and since @mpywell has opened a PR to implement this one, which is nearing completion, I believe we can close this open PR and refer to #283 for further updates.

Thanks @LethalServant for the original proposal and implementation, sorry it didn't get to the point where we could merge it, but hopefully the feature will be part of the plugin soon :)

PR #77, opened 2022-04-05
[fix] packer ignores pool parameter

mizhka:
The plugin ignores "pool" parameter due to incorrect copying of parameter value
into ConfigQemu. For example, in case of clone-vm for pooled template it results
into excluding of VM from pool.

This patch set "pool" parameter of ConfigQemu if parameter is not empty.

### DIFF
--- a/builder/proxmox/common/step_start_vm.go
+++ b/builder/proxmox/common/step_start_vm.go
@@ -87,6 +87,7 @@ func (s *stepStartVM) Run(ctx context.Context, state multistep.StateBag) multist
 		vmRef.SetNode(c.Node)
 		if c.Pool != "" {
 			vmRef.SetPool(c.Pool)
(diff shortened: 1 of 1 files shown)


carlpett (approved):
Nice catch, thanks!

nywilken (approved):
Nice. LGTM

mizhka:
@nywilken , @carlpett , thank you for quick response! I really appreciate your support.
Waiting for next release to enjoy packer with Proxmox pools.

nywilken:
Hi @mizhka happy to support. Thanks for the contributions. I'll be making a release of the plugin today. If you are not using HCL2 templates already I recommend looking into the hcl2_upgrade command and the [required_plugins](https://www.packer.io/docs/commands/init#init-command) block to help with installing and using new plugin releases as they become available.

PR #78, opened 2022-04-05
[debug] show REST API calls if packer is executed with "-debug" option

mizhka:
This patch allows to trace all REST API calls if packer is executed with "-debug"
option. Before this patch, there is no way to enable such tracing. It simplifies
analysis of packer-proxmox interaction and identify root cause of breaks.


nywilken (approved):
LGTM

PR #90, opened 2022-05-26
Add Secure Boot support

alex-sector:
EFI configuration have been updated to support Secure Boot feature merged in proxmox-api-go. The new syntax allows to create an EFI disk with default keys enrolled (which is by default in proxmox). 

Thus, Secure Boot will be available out-of-the-box without any tricks in the OS installation process.


lbajolet-hashicorp:
Since this was superseded by #130, I think we can close this PR, feel free to reopen if there's anything I missed here

PR #116, opened 2022-10-27
Update index.mdx

derrick-anderson:
Correct base path for hyperlink.

During the move to `developer.hashicorp`, the path structure for docs is no longer `/docs`.  I'm putting in this PR for a plugin I use. This change likely needs to be made across all plugin pages.


nywilken:
Thanks for the fix. It looks like the fix was addressed in #118, as well. I am going to close this change but If you happen to come across another issue please feel free to contribute and we will gladly review. Cheers!

PR #132, opened 2022-12-11
[feature] adds serial ports for iso and clone builder

### LINKED ISSUE #41: Allow adding serial sockets to template
#### Description

We can add "serial0" as the vga display type, but as far as I can tell there isn't any way to create serial0 in packer.

#### Use Case(s)

This would allow configurations of serial ports for the template. Proxmox allows port 0-3 it seems.

#### Potential configuration

```
serial_port [0, 1, 3]
```

#### Potential References
Proxmox has a page about Serial Terminal here: https://pve.proxmox.com/wiki/Serial_Terminal

mabeett:
Adds `serials` option which allows user adding from 1 up to 4 serials ports to the packer VM.

Closes #41

### DIFF
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -233,6 +233,12 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 	if len(c.Serials) > 4 {
 	    errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("serials admit up to 4 elements"))
 	}
+	res := regexp.MustCompile(`^(/dev/.+|socket)$`)
+	for idx := range c.Serials {
+		if !res.MatchString(c.Serials[idx]) {
+			errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("serials must respond to pattern \"/dev/.+\" or be \"socket\". It was \"%s\"", c.Serials[idx]))
+		}
+	}
 	if c.SCSIController == "" {
 		log.Printf("SCSI controller not set, using default 'lsi'")
 		c.SCSIController = "lsi"
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -230,6 +230,9 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 			errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("disk format must be specified for pool type %q", c.Disks[idx].StoragePoolType))
 		}
 	}
+	if len(c.Serials) > 4 {
+	    errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("serials admit up to 4 elements"))
+	}
 	if c.SCSIController == "" {
 		log.Printf("SCSI controller not set, using default 'lsi'")
 		c.SCSIController = "lsi"
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -58,6 +58,7 @@ type Config struct {
 	VGA            vgaConfig      `mapstructure:"vga"`
 	NICs           []nicConfig    `mapstructure:"network_adapters"`
 	Disks          []diskConfig   `mapstructure:"disks"`
+	Serials        []string       `mapstructure:"serials"`
 	Agent          config.Trilean `mapstructure:"qemu_agent"`
 	SCSIController string         `mapstructure:"scsi_controller"`
 	Onboot         bool           `mapstructure:"onboot"`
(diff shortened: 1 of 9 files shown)


mabeett:
> Just an FYI @mabeett  - "elements" is misspelt.

Thanks for your feedback!

I've just corrected the typo. If you consider there could be a better writing for the doc just let me know. English is not my mother's tongue.

builder/proxmox/common/config.go
```diff
@@ -57,6 +57,7 @@ type Config struct {
 	VGA            vgaConfig      `mapstructure:"vga"`
 	NICs           []nicConfig    `mapstructure:"network_adapters"`
 	Disks          []diskConfig   `mapstructure:"disks"`
+	Serials        []string       `mapstructure:"serials"`
```
lbajolet-hashicorp, line 60: Given the limitations described in the documentation, we can probably add a few checks in the `Prepare` step for the plugin, so we ensure that only valid values are accepted before we attempt to create the devices.

Like ensuring there are 4 or less elements, and that their values are the expected shape (assuming this is relevant?)

Also I wonder, why the limit of 4? Is it something enforced by the API?

lbajolet-hashicorp (changes requested):
The code looks OK to me overall, once the checks are implemented (or not, if they're not relevant), we can consider merging this.

Thanks @mabeett!

builder/proxmox/common/config.go
mabeett: Yes, The limit of 4 is imposed by the API. 
In case of using more than 4  the build fails with `Error creating VM: 400 Parameter verification failed` message.
During the next days I will work on your suggestion during my available spare time.

builder/proxmox/common/config.go
```diff
@@ -229,6 +230,15 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 			errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("disk format must be specified for pool type %q", c.Disks[idx].StoragePoolType))
 		}
 	}
+	if len(c.Serials) > 4 {
+	    errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("serials admit up to 4 elements"))
+	}
+	res := regexp.MustCompile(`^(/dev/.+|socket)$`)
+	for idx := range c.Serials {
```
lbajolet-hashicorp, line 237: If you have no use for the index in any place, you can probably change this loop to
```suggestion
	for _, serial := range c.Serials {
```

This way `serial` will be each entry in the `c.Serials` array, and you don't have to access it through the index in the following block.

builder/proxmox/common/config.go
```diff
@@ -229,6 +230,15 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 			errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("disk format must be specified for pool type %q", c.Disks[idx].StoragePoolType))
 		}
 	}
+	if len(c.Serials) > 4 {
+	    errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("serials admit up to 4 elements"))
```
lbajolet-hashicorp, line 234: Just a nit to rephrase the error here, so that users have a better idea what's happening with their serials.

```suggestion
	    errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("too many serials: %d serials defined, but proxmox accepts 4 elements maximum", len(c.Serials)))
```

lbajolet-hashicorp (approved):
Besides the last few nits I left, the code looks good to me, once addressed we can merge.

Pre-approving for later

builder/proxmox/common/config.go
mabeett: I've just pushed 2 commits and rebased the branch.
From my side I tested the changes manually the validation.

lbajolet-hashicorp: Thanks @mabeett, I left a review at the same time, just a few nits to address, regarding testing I think if we could validate at least the `Prepare` stuff automatically that'd be best.
Maybe consider adding this to `builder/proxmox/common/config_test.go`? Other than that, we can merge this, if you validated manually that's a good start.

lbajolet-hashicorp:
Hi @mabeett,

I took the liberty of adding one commit on your branch to test that the configs work as advertised, locally it does pass, we'll just wait until CI confirms it, and we can merge this PR.

mabeett:
Hi @lbajolet-hashicorp !, 

> I took the liberty of adding one commit on your branch to test that the configs work as advertised, locally it does pass, we'll just wait until CI confirms it, and we can merge this PR.

Thank you so much, 
my interventions on this repository are 100% best effort mostly 0% Clang knowledge. 
I'll take the commit as reference in case of future proposals. 

kind regards,

PR #139, opened 2023-01-17
Remove existing Cloud-Init drive before creation

### LINKED ISSUE #112: proxmox-clone will always fill the cloud-init drive with the ssh communicator config
When filing a bug, please include the following headings if possible. Any
example text in this template can be deleted.

#### Overview of the Issue

when using the clone builder, packer will always fill the cloud-init drive with informations from the ssh communicator, even when cloud-init is disabled. These writes will persist on clones of the templates (again, even if cloud init is disabled and the template doesn't have a c-i drive)

#### Reproduction Steps

make a basic proxmox-clone build, fill in ssh_username and ssh_private_key_file for the ssh communicator
plugin v1.0.8
packer v1.8.3

```
source "proxmox-clone" "ubuntu-2204-nomad-client" {
  // proxmox cluster settings
  proxmox_url = "${var.proxmox_api_url}"
  username = "${var.proxmox_api_token_id}"
  token = "${var.proxmox_api_token_secret}"
  insecure_skip_tls_verify = true

  // clone settings
  clone_vm = "ubuntu-2204-base"
  full_clone = true
    
  // vm general settings
  node = "neon"
  vm_id = "9005"
  vm_name = "ubuntu-2204-nomad-client"
  template_description = "Managed by Packer"

  // vm template settings
  cores = "1"
  memory = "1024" 
  qemu_agent = true

  // vm network settings
  network_adapters {
      model = "virtio"
      bridge = "vmbr0"
      firewall = "false"
      vlan_tag = "40"
  }

  // vm cloud-init settings
  cloud_init = false

  // packer ssh+http settings
  http_directory = "http" 
  ssh_username = "ansible"
  ssh_private_key_file = "/home/lanson/.ssh/id_packer"
  ssh_timeout = "20m"
}
```
something like that will cause the issue, the template that is being cloned here doesn't have a cloud-init drive, nor does it have the config to use one under `/etc/cloud/cloud.cfg.d/99-pve.cfg`

sebastian-de:
If a template is created from a VM with a Cloud-Init drive, the existing drive is neither deleted nor updated in stepFinalizeTemplateConfig and Proxmox throws `TASK ERROR: zfs error: cannot create 'rpool/data/vm-106-cloudinit': dataset already exists` and the build fails.

This PR adds `stepRemoveCloudInitDrive`, which will remove the Cloud-Init drive before finalizing the template.

To produce a clean template, also remove Cloud-Init configuration parameters set during build.
If `cloud_init` is set to true, the resulting drive will be empty.

Tested on PVE 7.3

Closes #112

### DIFF
--- a/builder/proxmox/common/step_remove_cloud_init_drive.go
+++ b/builder/proxmox/common/step_remove_cloud_init_drive.go
@@ -1,0 +1,91 @@
+package proxmox
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/Telmate/proxmox-api-go/proxmox"
+	"github.com/hashicorp/packer-plugin-sdk/multistep"
+	packersdk "github.com/hashicorp/packer-plugin-sdk/packer"
+)
+
+// stepRemoveCloudInitDrive removes the cloud-init cdrom and deletes
+// VM config parameters related to cloud-init.
+type stepRemoveCloudInitDrive struct{}
+
+type CloudInitDriveRemover interface {
+	GetVmConfig(*proxmox.VmRef) (map[string]interface{}, error)
+	SetVmConfig(*proxmox.VmRef, map[string]interface{}) (interface{}, error)
+}
+
+var _ CloudInitDriveRemover = &proxmox.Client{}
+
+func (s *stepRemoveCloudInitDrive) Run(ctx context.Context, state multistep.StateBag) multistep.StepAction {
+	ui := state.Get("ui").(packersdk.Ui)
+	client := state.Get("proxmoxClient").(CloudInitDriveRemover)
+	vmRef := state.Get("vmRef").(*proxmox.VmRef)
+
+	changes := make(map[string]interface{})
+	delete := []string{}
+
+	vmParams, err := client.GetVmConfig(vmRef)
+	if err != nil {
+		err := fmt.Errorf("error fetching template config: %s", err)
+		state.Put("error", err)
+		ui.Error(err.Error())
+		return multistep.ActionHalt
+	}
+	// The cloud-init drive is usually attached to ide0, but check the other controllers as well
+	diskControllers := []string{"ide0", "ide1", "ide2", "ide3"}
+	// Proxmox supports up to 6 SATA controllers (0 - 5)
+	for i := 0; i < 6; i++ {
+		sataController := fmt.Sprintf("sata%d", i)
+		diskControllers = append(diskControllers, sataController)
+	}
+	// and up to 31 SCSI controllers (0 - 30)
+	for i := 0; i < 31; i++ {
+		scsiController := fmt.Sprintf("scsi%d", i)
+		diskControllers = append(diskControllers, scsiController)
+	}
+
+	for _, controller := range diskControllers {
+		if vmParams[controller] != nil && strings.Contains(vmParams[controller].(string), "-cloudinit,media=cdrom") {
+			delete = append(delete, controller)
+		}
+	}
+
+	CloudInitParameters := []string{
+		"cipassword",
+		"ciuser",
+		"nameserver",
+		"searchdomain",
+		"sshkeys",
+	}
+	for i := 0; i < 16; i++ {
+		ipconfig := fmt.Sprintf("ipconfig%d", i)
+		CloudInitParameters = append(CloudInitParameters, ipconfig)
+	}
+	for _, parameter := range CloudInitParameters {
+		if vmParams[parameter] != nil {
+			delete = append(delete, parameter)
+		}
+	}
+
+	if len(delete) > 0 {
+		changes["delete"] = strings.Join(delete, ",")
+
+		_, err := client.SetVmConfig(vmRef, changes)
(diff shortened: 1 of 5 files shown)


builder/proxmox/common/step_remove_cloud_init_drive_test.go
```diff
@@ -0,0 +1,106 @@
...
+		t.Run(c.name, func(t *testing.T) {
+			remover := CloudInitDriveRemoverMock{
+				getConfig: func() (map[string]interface{}, error) {
+					return c.initialVMConfig, c.getConfigErr
+				},
+				setConfig: func(cfg map[string]interface{}) (string, error) {
+					if !c.expectCallSetConfig {
+						t.Error("Did not expect SetVmConfig to be called")
+					}
+					for key, val := range c.expectedVMConfig {
```
lbajolet-hashicorp, line 83: I'm confused by this test, what is it working with? Looking at the code the payload should essentially contain only a `delete` key/val, therefore the other values are nil by default, since no value exists for each key in the `expectedVMConfig`, unless I'm missing something?

lbajolet-hashicorp (changes requested):
Hi @sebastian-de,

Just left a few comments on the code and tests, I'm puzzled by the test more than the code.
Once you addressed the comments I left, I'll do another pass and if we're good, we'll merge that.

Thanks!

builder/proxmox/common/step_remove_cloud_init_drive.go
```diff
@@ -0,0 +1,81 @@
...
+
+	vmParams, err := client.GetVmConfig(vmRef)
+	if err != nil {
+		err := fmt.Errorf("error fetching template config: %s", err)
+		state.Put("error", err)
+		ui.Error(err.Error())
+		return multistep.ActionHalt
+	}
+	// The cloud-init drive is usually attached to ide0, but check the other controllers as well
+	ideControllers := []string{"ide0", "ide1", "ide2", "ide3"}
```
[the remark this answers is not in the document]
sebastian-de: Unfortunately it's not possible to link to the config entries directly, so I'm quoting https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes/{node}/qemu/{vmid}/config here:

> ide[n] string [...] Use volume as IDE hard disk or CD-ROM (n is 0 to 3).

Proxmox always defaults to IDE for Cloud-Init drives, but yes, in theory you can attach it to other controllers as well. I added them to the code.

builder/proxmox/common/step_remove_cloud_init_drive_test.go
sebastian-de: Oh dear, sorry for the headaches...
I rewrote the tests, don't know what I was thinking, when I wrote them initially.

builder/proxmox/common/step_remove_cloud_init_drive.go
```diff
@@ -0,0 +1,92 @@
...
+	// The cloud-init drive is usually attached to ide0, but check the other controllers as well
+	diskControllers := []string{"ide0", "ide1", "ide2", "ide3"}
+	// Proxmox supports up to 6 SATA controllers (0 - 5)
+	for i := 0; i < 6; i++ {
+		sataController := "sata" + strconv.Itoa(i)
+		diskControllers = append(diskControllers, sataController)
+	}
+	// and up to 31 SCSI controllers (0 - 30)
+	for i := 0; i < 31; i++ {
+		scsiController := "scsi" + strconv.Itoa(i)
```
lbajolet-hashicorp, line 49: Small nit: for such concats maybe a `Sprintf` would be a better option here
```suggestion
		scsiController := fmt.Sprintf("scsi%d", i)
```

builder/proxmox/common/step_remove_cloud_init_drive.go
```diff
@@ -0,0 +1,92 @@
...
+		err := fmt.Errorf("error fetching template config: %s", err)
+		state.Put("error", err)
+		ui.Error(err.Error())
+		return multistep.ActionHalt
+	}
+	// The cloud-init drive is usually attached to ide0, but check the other controllers as well
+	diskControllers := []string{"ide0", "ide1", "ide2", "ide3"}
+	// Proxmox supports up to 6 SATA controllers (0 - 5)
+	for i := 0; i < 6; i++ {
+		sataController := "sata" + strconv.Itoa(i)
```
lbajolet-hashicorp, line 44: Same comment here regarding the Sprintf
```suggestion
		sataController := fmt.Sprintf("sata%d", i)
```

lbajolet-hashicorp (approved):
Hi @sebastian-de,

Thanks for the reroll, the test makes more sense to me with the rewrite, I'm approving this PR.

I left a couple more nits to address, once you've had a chance to address them, we can merge.

builder/proxmox/common/step_remove_cloud_init_drive.go
sebastian-de: Done - also for the `ipconfig` case. Thanks again for your review!

PR #142, opened 2023-01-22
clone: Add Cloud-Init networking options

### LINKED ISSUE #7: [Proxmox-clone] add cloud-init config
_This issue was originally opened by @RemiDesgrange as hashicorp/packer#10251. It was migrated here as a result of the [Packer plugin split](https://github.com/hashicorp/packer/issues/8610#issuecomment-770034737). The original body of the issue is below._

<hr>

#### Community Note

#### Description

Proxmox clone could have some key to setup cloud-init (user, ssh key etc...)

#### Use Case(s)

Being able to override cloud-init base template we inherit from.

#### Potential configuration

```
"cloud_init": {
  "user": "user",
  "password": "pass",
  "ssh_keys": "XXXX"
  "ip_config": "ip=dhcp",
  "dns_domain": "example.com",
  "dns_servers": "1.1.1.1"
}
```

### LINKED ISSUE #9: Add ipconfig[n] support for Proxmox builder
_This issue was originally opened by @Weilbyte as hashicorp/packer#10169. It was migrated here as a result of the [Packer plugin split](https://github.com/hashicorp/packer/issues/8610#issuecomment-770034737). The original body of the issue is below._

<hr>

#### Community Note

#### Description

Add support for setting the cloud-init ipconfig setting when using the Proxmox builder. It should be found here as available config type => https://pve.proxmox.com/pve-docs/api-viewer/#/nodes/{node}/qemu/{vmid}/config

#### Use Case(s)

Easily setting up the VM template to use DHCP without having to mess around with the OS-specific configuration tools.

#### Potential configuration

For example setting v4/v6 to DHCP:
```
{
 "ipconfig": [
    id: 0,
    v4: "dhcp",
    v6: "dhcp"
  ]
}
```

sebastian-de:
This PR adds support for Cloud-Init networking options to the `proxmox-clone` builder.

This can become sightly messy if there are multiple network interfaces configured, since the order of `network_adapters` and `ipconfig` have to match.
I also wrote a [draft-PR](https://github.com/hashicorp/packer-plugin-proxmox/pull/141) to show another approach by extending `nicConfig` from the common config. This has the advantage that `ipconfig` can be set inside every NIC block.

But since Cloud-Init options are only available for the clone builder I tend to prefer the solution shown here. I think that very few people will define more than one network interface for the Packer run anyway.

But if desired, I can continue with the nested approach.

To be able to write proper tests, I had to convert `nicConfig` from the common config to uppercase, I'm not completely sure if this is the right approach.

Closes #7
Closes #9


builder/proxmox/clone/config.go
```diff
@@ -11,8 +17,19 @@ import (
...
-	CloneVM   string         `mapstructure:"clone_vm"`
+	CloneVM   string         `mapstructure:"clone_vm" required:"true"`
 	FullClone config.Trilean `mapstructure:"full_clone" required:"false"`
+
+	Nameserver   string              `mapstructure:"nameserver" required:"false"`
+	Searchdomain string              `mapstructure:"searchdomain" required:"false"`
+	Ipconfigs    []cloudInitIpconfig `mapstructure:"ipconfig" required:"false"`
+}
+
+type cloudInitIpconfig struct {
```
lbajolet-hashicorp, line 28: On a side-note, if this is linked to the `NICConfig`, is there a reason why this is specific to the `clone` builder? I wonder if that could apply to the `ISO` builder as well.

lbajolet-hashicorp (changes requested):
Hi @sebastian-de,

I left a couple nits and a few questions on the code, I'll let you address those and I can come back for a second round of review.

Aside from what I point out, the code looks good to me, thanks for the PR!

builder/proxmox/clone/config.go
```diff
@@ -11,8 +17,19 @@ import (
 type Config struct {
 	proxmox.Config `mapstructure:",squash"`
 
-	CloneVM   string         `mapstructure:"clone_vm"`
+	CloneVM   string         `mapstructure:"clone_vm" required:"true"`
```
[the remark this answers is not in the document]
sebastian-de: Yes, the build fails without `clone_vm`, so I guess this was forgotten initially.

builder/proxmox/clone/config.go
sebastian-de: At least right now, it's only supported for `clone` API-wise: https://github.com/Telmate/proxmox-api-go/blob/master/proxmox/config_qemu.go#L92

sebastian-de:
I addressed all your comments, thank you for your helpful review again!

And I forgot to mention: Tested on PVE 7.3

lbajolet-hashicorp (approved):
Hi @sebastian-de,

Thanks for the reroll, it's unfortunate that the ipconfig cannot be used for ISO in addition to clone, but if that's a limitation from the API, there's not much we can do from this side.

I'm merging this now, thanks again for the contribution!

PR #148, opened 2023-01-25
Download ISO files directly to PVE node

### LINKED ISSUE #49: promox-iso check if iso already exists on datastore
_This issue was originally opened by @Syntax3rror404 in https://github.com/hashicorp/packer/issues/10756 and has been migrated to this repository. The original issue description is below._

---
#### Community Note

#### Description

The iso is uploaded to the proxmox datastore with every Packer build. Since Packer already knows the checksum, a check would be possible. This feature was also recently implemented for the vsphere-iso provider. Please also implement this feature for the proxmox-iso provider.

#### Use Case(s)

-Saves time
-Building from the other network is possible (It takes forever to have to upload an iso from the Internet every time )
-Can protect the SSD from senseless writing 
-Better quality of the Packer Provider code 

#### Potential configuration
See vmware-iso provider
All variables are already implementent.

#### Potential References
See vmware-iso provider

### LINKED ISSUE #79: Do not upload an existing iso_url
When we set the `iso_url` property, the `proxmox-iso` builder always uploads the iso to the specified `iso_storage_pool` storage.

But this wastes time and wears out the storage.

This builder should only upload when the iso file does not exist at the given storage location.

sebastian-de:
Currently ISO files are downloaded by Packer and then uploaded to Proxmox in a second step.

Since Packer doesn't know the checksums of the files already on the Proxmox node, the ISO will get transferred regardless if it already exists. See the [upstream issue](https://bugzilla.proxmox.com/show_bug.cgi?id=3505) on why Proxmox doesn't provide checksums via it's API.

This PR introduces a new step that downloads the ISO files directly to the PVE node. This offers the benefits that double transfers (download to Packer host, then upload to PVE) are avoided and lets the PVE compare the checksums. If the checksums match, the downsload is skipped.
The only user-facing downside is, that we don't have the nice progress-bar the current approach offers.

I added an `iso_packer_upload` attribute to switch to old upload method if this should be desired. But I'm open to removing the old method completely.

Since `proxmox.DownloadIsoFromUrl()` is a new way to access `proxmox-api-go`, I don't see myself able to write tests for `stepDownloadISO`. I'm pretty sure that there's a straightforward way to do this, but with my limited Go knowledge I couldn't come up with one.

Tested on PVE 7.3.

Closes #49 
Closes #79


sebastian-de:
Further Information:

- API endpoint: https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes/{node}/storage/{storage}/download-url
- `proxmox-api-go` implementation: https://github.com/Telmate/proxmox-api-go/blob/3b29e08d4a97ae1f039dea99eef962bfb438ee70/proxmox/content_iso.go

sebastian-de:
Having a second look at this, I could call `client.PostWithTask()` directly - `proxmox.DownloadIsoFromUrl()` doesn't do anything else anyway.

On the other hand there are plans to split out more functions out of `proxmox.Client` (see [this issue](https://github.com/Telmate/proxmox-api-go/issues/210)), so it might become useful if we already have a proper way to test code like this.

builder/proxmox/iso/step_download_iso.go
```diff
@@ -0,0 +1,50 @@
...
+	"fmt"
+
+	"github.com/Telmate/proxmox-api-go/proxmox"
+	"github.com/hashicorp/packer-plugin-sdk/multistep"
+	packersdk "github.com/hashicorp/packer-plugin-sdk/packer"
+)
+
+// stepDownloadISO downloads an ISO file directly to the specified PVE node.
+// Checksums are also calculated and compared on the PVE node, not by Packer.
+type stepDownloadISO struct{}
```
lbajolet-hashicorp, line 14: The code looks good here, I'm just a little bit worried the name may be confusing vs. the `StepDownload` that comes with the sdk, we could probably rename it to make it clearer that the download will be performed exclusively on the PVE server, and not through go-getter here.
Would `stepDownloadISOOnPVE` be OK to you?

docs/builders/iso.mdx
```diff
@@ -82,6 +83,10 @@ in the image's Cloud-Init settings for provisioning.
 
 ### Optional:
 
+- `iso_packer_upload` (bool) - Download the specified `iso_url` first and
```
lbajolet-hashicorp, line 86: Because this changes the default behaviour here, I would proceed cautiously at first, I don't know if this may be problematic, but just in case what would you think about reversing this condition so the default behaviour is still using Packer's Download/UploadISO steps?
Something like `iso_download_pve`?

lbajolet-hashicorp (commented):
Hi @sebastian-de,

Thanks for the PR, I left a couple comments on the code, one naming issue where I think the added step may have a confusing name at first glance, so I think we can rename it to make it clearer what the intent is. The second comments pertains to the current behaviour being changed, while I think this won't be problematic, I think that in order to avoid surprises, we should strive not to change it directly, hence my suggestion to consciously select the PVE direct download.

Feel free to address those when you have time, I'll do a second pass on this PR after that.

builder/proxmox/iso/step_download_iso.go
sebastian-de: Sure, I renamed the step.

docs/builders/iso.mdx
sebastian-de: I guess you're right. Let's see if there arise any problems with the new method first.

docs/builders/iso.mdx
```diff
@@ -82,6 +82,9 @@ in the image's Cloud-Init settings for provisioning.
 
 ### Optional:
 
+- `iso_download_pve` (bool) - Download the specified `iso_url` directly to
+  the PVE node. Defaults to `false`
```
lbajolet-hashicorp, line 86: If you don't mind me chiming in for this, I think the comment might be slightly confusing with respects to the `to the PVE node`, my understanding is that Packer delegates downloading the ISO to the server, so in this case the ISO would be downloaded directly from the server, and Packer only gives the URI to get the ISO from.

I also think that users unfamiliar with the current behaviour may have a hard time deciding if this option is for them based on this information only, I think we should explain a bit more why that option exists and what it implies compared to the default behaviour.

I'll suggest a rephrasing, feel free to adapt it however you see fit.
```
- `iso_download_pve` (bool) - Download the specified `iso_url` directly from
  the PVE node. Defaults to `false`.
  By default Packer downloads the ISO and uploads it in a second step, this
  option lets Proxmox handle downloading the ISO directly from the server.
```

lbajolet-hashicorp (approved):
Hi @sebastian-de,

Thanks for applying the changes, I did another pass and left a suggestion regarding the docs on the new option, I think it can benefit from more explanations as to what changes when it is enabled.

Other than that, LGTM!
Pre-approving, let me know when you have decided what to do about my latest comment, and we can merge this PR.

docs/builders/iso.mdx
sebastian-de: Thanks, as a non-native speaker I sometimes have a hard time to decide between _downloading from/to/onto_.
I guess your version makes it all more clear.

lbajolet-hashicorp: Kein problem, happens to all of us!
Thanks for making the edit, I'll merge this PR as-is.

Issue #154, opened 2023-02-10
Builder Proxmox-iso error 506


ntatschner:
#### Overview of the Issue

Building a proxmox template with ubuntu 22.04 sourced from URL I'm getting and error 506 when attempting the build, packer file validates fine.

### Plugin and Packer version

From `proxmox 1.1.1` and packer latest

### Simplified Packer Buildfile

```
source "proxmox-iso" "ubuntu2204" {
  vm_name                  = "ubuntu2204templatebuilder"
  os                       = "l26"
  onboot                   = true
  cores                    = var.cores
  memory                   = var.memory
  sockets                  = var.sockets
  machine                  = "i440fx"
  proxmox_url              = var.proxmox_url
  username                 = var.proxmox_username
  token                    = var.token
  insecure_skip_tls_verify = true
  node                     = var.node_name
  template_name            = var.template_name
  ssh_username             = var.proxmox_builder_ssh_username
  ssh_password             = var.proxmox_builder_ssh_password
  ssh_timeout              = "20m"
  ssh_pty                  = true
  ssh_handshake_attempts   = 20
  #qemu_agent               = true
  #cloud_init               = true
  scsi_controller      = "virtio-scsi-pci"
  unmount_iso          = true
  template_description = "packer generated ubuntu-22.04.server-amd64"
  disks {
    type              = "virtio"
    disk_size         = var.disk_size
    storage_pool      = var.storage_pool
    storage_pool_type = "lvm-thin"
    format            = "qcow2"
  }
  iso_url          = var.iso_url
  iso_checksum     = var.iso_checksum
  iso_storage_pool = "local"
  boot_wait        = "10s"
  http_directory   = var.http_directory
  network_adapters {
    model    = "virtio"
    bridge   = "vmbr1"
    firewall = false
  }
  # cd_files = ["./${var.http_directory}/meta-data", "./${var.http_directory}/user-data"]
  # cd_label = "cidata"
  boot_command = [
    "<esc><wait><esc><wait><f6><wait><esc><wait>",
    "<bs><bs><bs><bs><bs>",
    "ip=${cidrhost("10.20.80.0/23", 9)}::${cidrhost("10.20.80.0/23", 1)}:${cidrnetmask("10.20.80.0/23")}::::${cidrhost("10.20.80.0/23", 1)} ",
    " autoinstall ds=nocloud-net;s=http://{{ .HTTPIP }}:{{ .HTTPPort }}/http/ ",
[...]

PR #156, opened 2023-02-10
docs: fix machine attribute docs

lbajolet-hashicorp:
The machine attribute docs mistakenly documented the possible values as either i440fx or q35, while the code checked for pc, pc-i440fx or q35.

To make the documentation closer to what the code actually does, we update this section.

Reported by #154


nywilken (approved):
Nice catch!

PR #157, opened 2023-02-12
clone: Add support to clone VM by ID, check if VM IDs are in valid range.

sebastian-de:
The clone builder currently supports cloning VMs by name only. Add a new attribute `clone_vm_id` to be able to clone VMs by ID, too.

The valid VMID ranges for Proxmox are hard to find, the only reference I found was this: https://pve-devel.pve.proxmox.narkive.com/Pa6mH1OP/avoiding-vmid-reuse#post8
I confirmed these numbers by testing on a PVE 7.3, so I also added the range-check for `vm_id` in a second commit.


lbajolet-hashicorp (approved):
Hi @sebastian-de,

LGTM! Thanks for this.

PR #159, opened 2023-02-13
docs: update machine docs without i440fx

lbajolet-hashicorp:
The i440fx machine type is advertised as supported by the Proxmox API client incorrectly, as it is then rejected by proxmox since the i440fx machine type is only supported with a version identifier, which will then be rejected by the API client.

To avoid confusion in the docs, we remove the reference to i440fx, and only document the `pc` and `q35` values as they are both accepted by proxmox and the API client.

Related to #154


nywilken (approved):
LGTM

PR #165, opened 2023-02-24
network_adapters: Add mtu attribute

### LINKED ISSUE #164: Please add MTU-Option for NIC-Configurations
Hello everyone,
is it possible to add the MTU-Option for NIC-Configurations? We are stuggeling with creating Machines on a Proxmox tied to a SDN and therefor need a reduced MTU.
The Proxmox-UI allows for this when creating NICs.

Overall I suspect the requested feature implementation to be similar to this request: https://github.com/hashicorp/packer/issues/7856 
It resulted in the following PR: https://github.com/hashicorp/packer/pull/7862/

Best regards

sebastian-de:
Related documentation: https://pve.proxmox.com/pve-docs/pve-admin-guide.html#qm_configuration
Tested on PVE 7.3

Closes #164

### DIFF
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -91,6 +91,7 @@ type NICConfig struct {
 	Model        string `mapstructure:"model"`
 	PacketQueues int    `mapstructure:"packet_queues"`
 	MACAddress   string `mapstructure:"mac_address"`
+	MTU          int    `mapstructure:"mtu"`
 	Bridge       string `mapstructure:"bridge"`
 	VLANTag      string `mapstructure:"vlan_tag"`
 	Firewall     bool   `mapstructure:"firewall"`
@@ -279,6 +280,9 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 		}
 		if c.NICs[idx].Model != "virtio" && c.NICs[idx].PacketQueues > 0 {
 			errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("network_adapters[%d].packet_queues can only be set for 'virtio' driver", idx))
+		}
+		if (c.NICs[idx].MTU < 0) || (c.NICs[idx].MTU > 65520) {
+			errs = packersdk.MultiErrorAppend(errs, errors.New("network_adapters[%d].mtu only positive values up to 65520 are supported"))
 		}
 	}
 	for idx := range c.Disks {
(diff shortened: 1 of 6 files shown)


builder/proxmox/common/config.go
```diff
@@ -280,6 +281,9 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 		if c.NICs[idx].Model != "virtio" && c.NICs[idx].PacketQueues > 0 {
 			errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("network_adapters[%d].packet_queues can only be set for 'virtio' driver", idx))
 		}
+		if (c.NICs[idx].MTU < 0) || (c.NICs[idx].MTU > 65520) {
```
lbajolet-hashicorp, line 284: Given we are performing multiple checks here on each NIC, we could probably change the way we iterate on those by using a `for idx, nic := range c.NICs`, this way we can access the values without having to work with the slice directly.

lbajolet-hashicorp: This can be done in a separate commit/PR though, no need to amend the existing ones

lbajolet-hashicorp (approved):
Hi @sebastian-de,

The code looks good to me, I left a question and a remark, I'll let you decide if you want to fold the change in this PR, but aside from that, we should be ready to merge soon.

Thanks!

docs/builders/iso.mdx
```diff
@@ -159,6 +159,10 @@ in the image's Cloud-Init settings for provisioning.
     not set, defaults to a random MAC. If value is "repeatable", value of MAC
     address is deterministic based on VM ID and NIC ID.
 
+  - `mtu` (int) - Set the maximum transmission unit for the adapter. Valid
+    range: 0 - 65520. If set to `1`, the MTU is inherited from the bridge
+    the adapter is attached to. Defaults to `0` (use Proxmox default).
```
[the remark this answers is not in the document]
sebastian-de: I couldn't find anything on that in the Proxmox documentation. To be honest I don't even know if Proxmox is setting the default, or if this is done by QEMU once the VM starts. In my tests I always ended up with an MTU of 1500, when was `mtu` unset.
But since that is also the the MTU of the underlying bridge, the default _might_ also be to just use the MTU of the connected bridge.

builder/proxmox/common/config.go
sebastian-de: That's a great suggestion - I guess that would also make sense for at least some of the iterations on `c.Disks` and `c.AdditionalISOFiles` as well. If so, I can change these as well and open a separate PR on that.

lbajolet-hashicorp: Oh true, if we're to fix all of them, might as well do them in a later PR then; I'll merge this one in the meantime, feel free to come back to the loops later, thanks @sebastian-de.

PR #120, opened 2022-10-31
Fix names of registered builders

sebastian-de:
Fixes duplicate "proxmox-" prefixes in builder names when running the plugin as an external process.

Doesn't affect bundling by Packer since builder names are defined in [vendored_plugins.go](https://github.com/hashicorp/packer/blob/11e71729f187bad76d28f1701cf2c56e7b912c0f/command/vendored_plugins.go#L117
)

Closes #119

PR #175, opened 2023-03-13
Register "proxmox" builder again

sebastian-de:
This solves the issue described in https://github.com/hashicorp/packer/issues/12309

While updating the builder names in #120 I wasn't aware that we still need to register a builder with `plugin.DEFAULT_NAME`.

Apparently a lot of users still use the old builder name.

Closes [#12309](https://github.com/hashicorp/packer/issues/12309)


nywilken (approved):
Good catch @sebastian-de this was not expected. Thanks for linking out the historical info.

PR #183, opened 2023-04-06
Add iso_download_pve feature to additional_iso_files

ELT4N1N:
This PR adds the `iso_download_pve` feature to `additional_iso_files`. This is helpful for iso files that are a on the bigger side and are often used for multiple VMs, like the virtio drivers. The flag can only be used together with iso_url.
The code is based on the code from the iso module and shares it with it. 

It was tested on PVE  7.4-3.

### DIFF
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -759,7 +759,7 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 			errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("one of iso_file, iso_url, or a combination of cd_files and cd_content must be specified for AdditionalISO file %s", c.AdditionalISOFiles[idx].Device))
 		}
 		if len(c.AdditionalISOFiles[idx].ISOConfig.ISOUrls) == 0 && c.AdditionalISOFiles[idx].ISOConfig.RawSingleISOUrl == "" && c.AdditionalISOFiles[idx].ISODownloadPVE {
-			err = packersdk.MultiErrorAppend(errs, fmt.Errorf("iso_download_pve can only be used together with iso_url"))
+			errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("iso_download_pve can only be used together with iso_url"))
 		}
 	}
 	if c.EFIDisk != "" {
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -226,7 +226,10 @@ type additionalISOsConfig struct {
 	// Proxmox storage pool onto which to upload
 	// the ISO file.
 	ISOStoragePool string `mapstructure:"iso_storage_pool"`
-	ISODownloadPVE bool   `mapstructure:"iso_download_pve"`
+	// Download the ISO directly from the PVE node rather than through Packer.
+	//
+	// Defaults to `false`
+	ISODownloadPVE bool `mapstructure:"iso_download_pve"`
 	// If true, remove the mounted ISO from the template after finishing. Defaults to `false`.
 	Unmount              bool   `mapstructure:"unmount"`
 	ShouldUploadISO      bool   `mapstructure-to-hcl2:",skip"`
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -226,6 +226,7 @@ type additionalISOsConfig struct {
 	// Proxmox storage pool onto which to upload
 	// the ISO file.
 	ISOStoragePool string `mapstructure:"iso_storage_pool"`
+	ISODownloadPVE bool   `mapstructure:"iso_download_pve"`
 	// If true, remove the mounted ISO from the template after finishing. Defaults to `false`.
 	Unmount              bool   `mapstructure:"unmount"`
 	ShouldUploadISO      bool   `mapstructure-to-hcl2:",skip"`
@@ -756,6 +757,9 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 		}
 		if options != 1 {
 			errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("one of iso_file, iso_url, or a combination of cd_files and cd_content must be specified for AdditionalISO file %s", c.AdditionalISOFiles[idx].Device))
+		}
+		if len(c.AdditionalISOFiles[idx].ISOConfig.ISOUrls) == 0 && c.AdditionalISOFiles[idx].ISOConfig.RawSingleISOUrl == "" && c.AdditionalISOFiles[idx].ISODownloadPVE {
+			err = packersdk.MultiErrorAppend(errs, fmt.Errorf("iso_download_pve can only be used together with iso_url"))
 		}
 	}
 	if c.EFIDisk != "" {
--- a/builder/proxmox/common/step_download_iso_on_pve.go
+++ b/builder/proxmox/common/step_download_iso_on_pve.go
@@ -42,10 +42,7 @@ func Download_iso_on_pve(state multistep.StateBag, ISOUrls []string, ISOChecksum
 	for _, url := range ISOUrls {
 		var checksum string
 		var checksumType string
-		if ISOChecksum == "none" {
-			checksum = ""
-			checksumType = ""
-		} else {
+		if ISOChecksum != "none" {
 			gr := &getter.Request{
 				Src: url + "?checksum=" + ISOChecksum,
 			}
--- a/builder/proxmox/common/step_download_iso_on_pve.go
+++ b/builder/proxmox/common/step_download_iso_on_pve.go
@@ -25,6 +25,7 @@ func (s *stepDownloadISOOnPVE) Run(ctx context.Context, state multistep.StateBag
 	var isoStoragePath string
 	isoStoragePath = Download_iso_on_pve(state, s.ISO.ISOUrls, s.ISO.ISOChecksum, s.ISO.ISOStoragePool)
 
+	// If available, set the file path to the downloaded iso file on the node
 	if isoStoragePath != "" {
 		s.ISO.ISOFile = isoStoragePath
 		return multistep.ActionContinue
@@ -33,11 +34,20 @@ func (s *stepDownloadISOOnPVE) Run(ctx context.Context, state multistep.StateBag
 	return multistep.ActionHalt
 }
 
+// Download_iso_on_pve abstracts the checksum and download process os that the code can be shared between
+// the common module and the iso module. This is necessary because both handle the storage path to the iso differently.
+//
+// The function takes a list of URLs to download the iso and tries them one another.
+// If a download was successful it skips the additonal downlaod mirrors and returns the path to the iso on the node.
+//
+// Returns: When successful, the path to the iso on the node, else an empty string.
 func Download_iso_on_pve(state multistep.StateBag, ISOUrls []string, ISOChecksum string, ISOStoragePool string) string {
 	ui := state.Get("ui").(packersdk.Ui)
 	client := state.Get("proxmoxClient").(*proxmox.Client)
 	c := state.Get("config").(*Config)
 
+	// Generate ISOConfig configuration attributes in the format defined for packer-plugin-sdk
+	// and use go-getter to generate parameters compatible with the Proxmox-API.
 	var isoConfig proxmox.ConfigContent_Iso
 	for _, url := range ISOUrls {
 		var checksum string
@@ -68,6 +78,7 @@ func Download_iso_on_pve(state multistep.StateBag, ISOUrls []string, ISOChecksum
 
 		ui.Say(fmt.Sprintf("Beginning download of %s to node %s", isoConfig.DownloadUrl, isoConfig.Node))
 		err := proxmox.DownloadIsoFromUrl(client, isoConfig)
+		// On error continues with the next URL and logs the error
 		if err != nil {
 			state.Put("error", err)
 			ui.Error(err.Error())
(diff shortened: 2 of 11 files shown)


nywilken:
Thanks for opening up this change we will review as soon as we get a chance.

builder/proxmox/common/config.go
```diff
@@ -364,6 +365,9 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 		if options != 1 {
 			errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("one of iso_file, iso_url, or a combination of cd_files and cd_content must be specified for AdditionalISO file %s", c.AdditionalISOFiles[idx].Device))
 		}
+		if len(c.AdditionalISOFiles[idx].ISOConfig.ISOUrls) == 0 && c.AdditionalISOFiles[idx].ISOConfig.RawSingleISOUrl == "" && c.AdditionalISOFiles[idx].ISODownloadPVE {
+			err = packersdk.MultiErrorAppend(errs, fmt.Errorf("iso_download_pve can only be used together with iso_url"))
```
nywilken, line 369: It look like you want to append to errs here. 
```suggestion
			errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("iso_download_pve can only be used together with iso_url"))
```

ELT4N1N: Yes it should be errs

builder/proxmox/common/step_download_iso_on_pve.go
```diff
@@ -0,0 +1,87 @@
...
+
+	if isoStoragePath != "" {
+		s.ISO.ISOFile = isoStoragePath
+		return multistep.ActionContinue
+	}
+	// Abort if no ISO can be downloaded
+	return multistep.ActionHalt
+}
+
+func Download_iso_on_pve(state multistep.StateBag, ISOUrls []string, ISOChecksum string, ISOStoragePool string) string {
```
nywilken, line 36: It wasn't immediately clear to me  what this was returning. I recommend documenting this function to indicate what it does.

builder/proxmox/common/step_download_iso_on_pve.go
```diff
@@ -0,0 +1,87 @@
...
+			gc := getter.Client{}
+			fileChecksum, err := gc.GetChecksum(context.TODO(), gr)
+			if err != nil {
+				state.Put("error", err)
+				ui.Error(err.Error())
+				continue
+			}
+			checksum = hex.EncodeToString(fileChecksum.Value)
+			checksumType = fileChecksum.Type
+		}
```
nywilken, line 61: Seeing as both checksum and checksumType are already empty strings as defined on lines 43 and 44. I recommend inverting this if/else statement to drop the unnecessary else.  
```suggestion
		if ISOChecksum != "none" {
			gr := &getter.Request{
				Src: url + "?checksum=" + ISOChecksum,
			}
			gc := getter.Client{}
			fileChecksum, err := gc.GetChecksum(context.TODO(), gr)
			if err != nil {
				state.Put("error", err)
				ui.Error(err.Error())
				continue
			}
			checksum = hex.EncodeToString(fileChecksum.Value)
			checksumType = fileChecksum.Type
		}
```

builder/proxmox/common/step_download_iso_on_pve.go
```diff
@@ -0,0 +1,87 @@
...
+		if err != nil {
+			state.Put("error", err)
+			ui.Error(err.Error())
+			continue
+		}
+		isoStoragePath := fmt.Sprintf("%s:iso/%s", isoConfig.Storage, isoConfig.Filename)
+		ui.Say(fmt.Sprintf("Finished downloading %s", isoStoragePath))
+		return isoStoragePath
+	}
+	return ""
```
nywilken, line 83: Assuming if we it means no ISOUrls were specified, correct? 

If so I would recommend add a guard clause and return early before line 42

```go
if len(ISOUrls) == 0 {
 return ""
}
```

nywilken (commented):
I left a few more suggestions and pinged @lbajolet-hashicorp for another set of eyes as he has more insight into the workings of Proxmox

builder/proxmox/common/step_download_iso_on_pve.go
ELT4N1N: You are right. I will add some comments to make it more clear.

builder/proxmox/common/step_download_iso_on_pve.go
ELT4N1N: That's a good idea. I just took the code that was already there and didn't want to modify it too much. 
Didn't know how Go defines undefined strings. I'm not a Go developer :)

builder/proxmox/common/step_download_iso_on_pve.go
ELT4N1N: It returns an empty string, if no download was successful and theoretically if no URLs were given (which should not be possible because of the check in the config).
If ISOUrls would be empty, the for loop should do nothing and therefore directly jump to return "".

builder/proxmox/common/step_download_iso_on_pve.go
```diff
@@ -0,0 +1,87 @@
...
+		ui.Say(fmt.Sprintf("Beginning download of %s to node %s", isoConfig.DownloadUrl, isoConfig.Node))
+		err := proxmox.DownloadIsoFromUrl(client, isoConfig)
+		if err != nil {
+			state.Put("error", err)
+			ui.Error(err.Error())
+			continue
+		}
+		isoStoragePath := fmt.Sprintf("%s:iso/%s", isoConfig.Storage, isoConfig.Filename)
+		ui.Say(fmt.Sprintf("Finished downloading %s", isoStoragePath))
+		return isoStoragePath
```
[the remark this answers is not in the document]
ELT4N1N: So in my understanding ISOUrls is an array of URLs to the same iso file (with the same checksum), basically mirror hosts. If the first URL is down or the file does not match the checksum it tries the next URL and so on. 
Therefore, if the first download was successful we don't need to try the other ones and can just return the path to the iso.

At least that's how I understood the code. I didn't try that feature yet because it seems to only be available for HCL2 configs and mine are JSON.

builder/proxmox/common/step_download_iso_on_pve.go
lbajolet-hashicorp: In terms of code style, Go encourages usage of camelCase or PascalCase for function names depending on visibility, so I would suggest we rename this to `DownloadISOOnPVE` in order to be more in-line with the guidelines for the language.

builder/proxmox/iso/step_download_iso_on_pve.go
```diff
@@ -5,44 +5,32 @@ package proxmoxiso
...
-	"github.com/Telmate/proxmox-api-go/proxmox"
+	proxmoxcommon "github.com/hashicorp/packer-plugin-proxmox/builder/proxmox/common"
 	"github.com/hashicorp/packer-plugin-sdk/multistep"
-	packersdk "github.com/hashicorp/packer-plugin-sdk/packer"
 )
 
 // stepDownloadISOOnPVE downloads an ISO file directly to the specified PVE node.
 // Checksums are also calculated and compared on the PVE node, not by Packer.
-type stepDownloadISOOnPVE struct{}
+type stepDownloadISOOnPVE struct {
```
lbajolet-hashicorp, line 15: Is there a reason to keep this step in the code since we're defining a `common` step for this purpose? I would think the `common` implementation supersedes this one, and therefore we could remove this code.

builder/proxmox/common/step_download_iso_on_pve.go
lbajolet-hashicorp: In case nothing was downloaded, this should be an error, shouldn't it? I think the function would benefit from returning both a string and an error with context on what went wrong, so we can forward this to the user, and they can troubleshoot more effectively.

builder/proxmox/common/step_download_iso_on_pve.go
```diff
@@ -0,0 +1,97 @@
...
+	var isoStoragePath string
+	isoStoragePath = Download_iso_on_pve(state, s.ISO.ISOUrls, s.ISO.ISOChecksum, s.ISO.ISOStoragePool)
+
+	// If available, set the file path to the downloaded iso file on the node
+	if isoStoragePath != "" {
+		s.ISO.ISOFile = isoStoragePath
+		return multistep.ActionContinue
+	}
+	// Abort if no ISO can be downloaded
+	return multistep.ActionHalt
```
lbajolet-hashicorp, line 34: Halting is correct here, but I would argue we should present an error to the user with the `Ui`. You can get this from the state bag, it's available under the `ui` key, and its type is `packersdk.Ui`.
SEE: builder/proxmox/iso/step_upload_iso.go:28 for an example of how to get it from the state.

builder/proxmox/common/config.go
```diff
@@ -84,6 +84,7 @@ type additionalISOsConfig struct {
 	Device                string `mapstructure:"device"`
 	ISOFile               string `mapstructure:"iso_file"`
 	ISOStoragePool        string `mapstructure:"iso_storage_pool"`
+	ISODownloadPVE        bool   `mapstructure:"iso_download_pve"`
```
lbajolet-hashicorp, line 87: This would benefit from documentation to make it clear what the behaviour should be if defined.

builder/proxmox/common/step_download_iso_on_pve.go
```diff
@@ -0,0 +1,97 @@
...
+			Filename:          path.Base(url),
+			ChecksumAlgorithm: checksumType,
+			Checksum:          checksum,
+		}
+
+		ui.Say(fmt.Sprintf("Beginning download of %s to node %s", isoConfig.DownloadUrl, isoConfig.Node))
+		err := proxmox.DownloadIsoFromUrl(client, isoConfig)
+		// On error continues with the next URL and logs the error
+		if err != nil {
+			state.Put("error", err)
```
lbajolet-hashicorp, line 83: We should only set the error if all the attempts have failed, otherwise it's premature as subsequent tries may succeed.

builder/proxmox/common/step_download_iso_on_pve.go
```diff
@@ -0,0 +1,97 @@
...
+		isoStoragePath := fmt.Sprintf("%s:iso/%s", isoConfig.Storage, isoConfig.Filename)
+		ui.Say(fmt.Sprintf("Finished downloading %s", isoStoragePath))
+		// Returns the path to the iso on the node
+		return isoStoragePath
+	}
+	// Returns an empty string, which means download was not successful.
+	return ""
+}
+
+func (s *stepDownloadISOOnPVE) Cleanup(state multistep.StateBag) {
```
lbajolet-hashicorp, line 96: I wonder, should we clean the downloaded ISOs from the server? Is there an API call to do this?

lbajolet-hashicorp (changes requested):
Hi @ELT4N1N,

First off, apologies for the long turnaround time, I forgot to take a look at this when @nywilken mentioned me.

I took a look at the code, and left a couple of comments/questions which I think would benefit from being addressed before we merge this PR.

On those, one question that is in my opinion the most important is why do we have two separate steps for downloading those ISOs? I would think one step downloading all the ISOs would be enough for both the ISO for booting the image, and the other ISOs as well should there be more.

Other than that, the base code looks OK, there's room for improvement I think, but the framework for supporting this is in place, so now we can iterate on it.

Let me know if you need help on anything, I should be able to take a look at the code if need be.

Thanks for opening this PR!

ELT4N1N:
Thanks for your feedback. I'm not an developer so your inputs are helpful to me. I thought by keeping most of the code that is already in packer it would be easier to merge. I will look into your suggestions.

builder/proxmox/common/step_download_iso_on_pve.go
ELT4N1N: I can look into it, but I would suggest keeping it behind a flag that enables or disables the cleanup. I don't think we have to delete the e. g. virtio drivers every time as it can be reused in the next template. At least an my use case of installing multiple windows versions it is faster when keeping the files.

builder/proxmox/common/config.go
ELT4N1N: added it to "clone" documentation as it already is in the "iso" part

builder/proxmox/iso/step_download_iso_on_pve.go
ELT4N1N: The problem is that the file path is returned in two different ways. In `common` I can return it via the `additionalISOsConfig` struct. But in `iso` I can only get a string variable, so it was solved by putting it in the `iso-config` state bag.
It is possible to move that logic to `stepUploadISO` and 'stepUploadAdditionalISO' and reuse the `DownloadISOOnPVE` function there. I don't know if this is a better solution.

builder/proxmox/common/step_download_iso_on_pve.go
ELT4N1N: It seems to be possible but I think that would be a new feature as it is also not done for the normal upload of iso files. The only iso file that gets deleted is the auto-generated one in `stepUploadAdditionalISO`. So I think this could be one solution for all these procedures and not specific to this step.

ELT4N1N:
@lbajolet-hashicorp Sorry I overlooked your change request of the function renaming and now it seems to be a conflict that I can't solve form my side.

builder/proxmox/common/config.go
rgl: this variable name is somewhat ambiguous, can you please rename it to `iso_download_from_pve_node` (the `node` word is used to match the proxmox-iso source `node` property)?

lbajolet-hashicorp:
Hey @ELT4N1N,

Sorry it took so long for us to take another look at this PR, let's get the ball rolling to get it merged soon!

I took the liberty to fix the conflicts on the code, it should be building on top of the current main now.

I'll take another look at the code now, if I can't finish my review today, you can expect it to be done by tomorrow.

Thanks!

Issue #127, opened 2022-11-25
Ballooning


modem7:
#### Description

It would be great if we could set the min/max RAM available to the VM. 

Currently we can set the memory as a single value, but Proxmox allows for ballooning.

#### Potential configuration

```
Memory:
Minimum_Memory:
Shares:
Ballooning Device:
```

![image](https://user-images.githubusercontent.com/4349962/204045278-c08d38ca-7fbb-4c67-9e29-501d14fd1b87.png)

#### Potential References
https://pve.proxmox.com/wiki/Dynamic_Memory_Management

stfnx:
Absolutely agree. This would be super nice indeed!

sebastian-de:
I had a look at this over the last days. The relevant API options look simple at first, but implementing this for Packer is a bit more complicated.
These are the API options:
```
-balloon   <integer> (0 - N)

            Amount of target RAM for the VM in MB. Using zero disables the
            ballon driver.

-shares    <integer> (0 - 50000)   (default=1000)

            Amount of memory shares for auto-ballooning. The larger the
            number is, the more memory this VM gets. Number is relative to
            weights of all other running VMs. Using zero disables
            auto-ballooning. Auto-ballooning is done by pvestatd.
```
https://pve.proxmox.com/pve-docs/pve-admin-guide.html#qm_virtual_machines_settings states:

> Even when using a fixed memory size, the ballooning device gets added to the VM, because it delivers useful information such as how much memory the guest really uses. In general, you should leave ballooning enabled, but if you want to disable it (like for debugging purposes), simply uncheck Ballooning Device or set
> balloon: 0
> in the configuration.

Proxmox differentiates between `0` and unset variables. In case of the `balloon` option, _unset_ means to use the default (ballooning disabled, but balloon driver enabled), `0` means to disable the ballon driver completely.

Since an uninitialized Integer is always `0` in Go, not setting this option would alter the default behavior and deactivate the balloon driver which is something you almost never want to do.

`proxmox-api-go` seems to have a check in place to prevent that for `balloon`: https://github.com/Telmate/proxmox-api-go/blob/ad6090b8284272279e238bcaafadaa0e28536d7c/proxmox/config_qemu.go#L138

Same for `shares`: _unset_ means `1000`, `0` means to disable auto-ballooning, which again is something you probably don't want, if you have ballooning enabled.

So all in all I lean to the following implementation:
[...]

PR #186, opened 2023-04-13
Add numa attribute

sebastian-de:
This PR adds basic NUMA support to the builder as requested on the forum:

https://discuss.hashicorp.com/t/packer-proxmox-numa-support/51228

Tested on PVE 7.4.

### DIFF
--- a/builder/proxmox/common/step_start_vm.go
+++ b/builder/proxmox/common/step_start_vm.go
@@ -119,6 +119,7 @@ func (s *stepStartVM) Run(ctx context.Context, state multistep.StateBag) multist
 		Memory:       c.Memory,
 		QemuCores:    c.Cores,
 		QemuSockets:  c.Sockets,
+		QemuNuma:     &c.Numa,
 		QemuOs:       c.OS,
 		Bios:         c.BIOS,
 		EFIDisk:      generateProxmoxEfi(c.EFIConfig),
--- a/builder/proxmox/clone/config.hcl2spec.go
+++ b/builder/proxmox/clone/config.hcl2spec.go
@@ -94,6 +94,7 @@ type FlatConfig struct {
 	Cores                     *int                               `mapstructure:"cores" cty:"cores" hcl:"cores"`
 	CPUType                   *string                            `mapstructure:"cpu_type" cty:"cpu_type" hcl:"cpu_type"`
 	Sockets                   *int                               `mapstructure:"sockets" cty:"sockets" hcl:"sockets"`
+	Numa                      *bool                              `mapstructure:"numa" cty:"numa" hcl:"numa"`
 	OS                        *string                            `mapstructure:"os" cty:"os" hcl:"os"`
 	BIOS                      *string                            `mapstructure:"bios" cty:"bios" hcl:"bios"`
 	EFIConfig                 *proxmox.FlatefiConfig             `mapstructure:"efi_config" cty:"efi_config" hcl:"efi_config"`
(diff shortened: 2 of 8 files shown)


builder/proxmox/common/step_start_vm.go
```diff
@@ -119,6 +119,7 @@ func (s *stepStartVM) Run(ctx context.Context, state multistep.StateBag) multist
 		Memory:       c.Memory,
 		QemuCores:    c.Cores,
 		QemuSockets:  c.Sockets,
+		QemuNuma:     &c.Numa,
```
lbajolet-hashicorp, line 122: Is there a reason why this is a `*bool`? The rest of the arguments look like they're passed by copy so I wonder why this one in particular is different.

lbajolet-hashicorp (commented):
Hi @sebastian-de,

Sorry for taking so long to review this, looking at it now.

Fairly straightforward code I must say :) This looks good to me, I do wonder if there could be tests to ensure this is well forwarded to the API?

Other than that, I would maybe ask about the relevance of the option, on its own, how does the option manifests itself for the VM? With qemu my understanding is that you can setup a NUMA topology for the virtualised hardware, but it involves more than a flag to define it, so I wonder what this option effectively does in the end.

Besides that concern, this looks good to me, I'll swing back to approve this code once you've responded to this comment.

Thanks!

builder/proxmox/common/step_start_vm.go
sebastian-de: That's what `proxmox-api-go` does - we use their `ConfigQemu` struct directly.
They do this to have a way to differentiate between `nil` and `false`. The Proxmox-API is sometimes tricky to work with, since setting something to `false` (or `0` for integers) often means something different than the default. Having a `nil` pointer is usually the `proxmox-api-go` way of saying "attribute is unset, use the default".
See here for some details on this matter: https://github.com/hashicorp/packer-plugin-proxmox/issues/127#issuecomment-1464030102

sebastian-de:
Hey @lbajolet-hashicorp , thanks for your review!

> Sorry for taking so long to review this, looking at it now.

No problem!

> Fairly straightforward code I must say :) This looks good to me, I do wonder if there could be tests to ensure this is well forwarded to the API?

Sure, I can add a test for it.

> Other than that, I would maybe ask about the relevance of the option, on its own, how does the option manifests itself for the VM? With qemu my understanding is that you can setup a NUMA topology for the virtualised hardware, but it involves more than a flag to define it, so I wonder what this option effectively does in the end.

It's my understanding, that if NUMA on Proxmox is enabled without any further options, it tries to assign only cores from one physical socket to that VM and uses the host memory connected to that socket, see https://pve.proxmox.com/wiki/NUMA

builder/proxmox/common/step_start_vm.go
lbajolet-hashicorp: I see, in this case it seems that it will never be `nil` since the attribute is a non-pointer boolean in the configs, so it will always be a pointer to either `true` or `false` (default).
I presume that it shouldn't be a problem though, it just means that we'll never have the default (unspecified) value anymore for this attribute.
Is this correct?

lbajolet-hashicorp (approved):
All good then, thanks for the precision! In this case we can merge this soon to support enabling NUMA so Proxmox will be better able to pin resources for the VM.
I'll wait a bit regarding tests, once you've implemented those we can circle back and merge this PR.

Pre-approving since the code looks OK to me in its current state.

sebastian-de:
> I do wonder if there could be tests to ensure this is well forwarded to the API?

> Sure, I can add a test for it.

A vaguely remembered that it was difficult to write proper tests for the Proxmox builder and trying to implement a test for the numa attribute I now remember why.
This comment sums it up: https://github.com/hashicorp/packer-plugin-proxmox/blob/76fe4542920228c777fffa2437ed7fc79977abf8/builder/proxmox/common/step_start_vm_test.go#L152

So right now I don't see a way to properly check the `params` map that gets send to the API.

lbajolet-hashicorp:
Hi @sebastian-de,

I understand, it would be preferable if we could test it better, but if that's a headache to get it working, we can go with the change regardless, there's not a lot of changes here, and if you tested manually that it worked, that should be good enough for now.

Thanks for the update, I'll trigger the merge now.

sebastian-de:
@lbajolet-hashicorp as always: Thanks a lot for your great feedback!

PR #194, opened 2023-05-26
cleanup github workflows

lbajolet-hashicorp:
Remove the test-example.yml workflow as it is not necessary and contravenes the new guidelines for Github Actions.


nywilken (approved):
LGTM

PR #200, opened 2023-06-24
Remove extra line to fix formatting

dglinder:
The formatting of the 'cd_content' section of the documentation is broken on the MDX interpreter on the website - it looks like the extra blank line between the two paragraphs combined with the four spaces preceding the "-" make the MDX interpreter see the block as code and not Markdown coded text.


lbajolet-hashicorp (approved):
Hi @dglinder,

Thanks for this fix, locally tested to rebuild the docs without that extra line, and it does seem to fix the rendering problem.

Will merge this now

PR #203, opened 2023-07-10
use correct value for discard parameter when disabled

### LINKED ISSUE #202: incorrect flag values for controlling ssd trim behaviour
#### Overview of the Issue

#199 added support for a new disk flag `discard`, but I think the values may have been incorrectly specified. At least on Proxmox 7.4-13 the accepted values are `ignore` and `on`, but the code assumes it's `on` and `off`. Presumably, this is the case on older Proxmox setups, but I don't have access to one to confirm.

Latest [proxmox API docs](https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes/{node}/qemu/{vmid}/config) (search for `discard`) also have it as:
```
discard=<ignore | on> 
```

#### Reproduction Steps

Build an image using this config:

```hcl
source "proxmox-iso" "debian" {
  # ...
  disks {
    discard  = false
    # ...
```

The VM will fail to start with the following error:

<code>==> proxmox-iso.debian: Error creating VM: error creating VM: 400 Parameter verification failed., error status: {"errors":{"scsi0":"invalid format - format error\nscsi0.discard: value 'off' does not have a value in the enumeration 'ignore, on'\n"},"data":null}
</code>

### Plugin and Packer version

Packer v1.9.1

jeinwag:
Fixes #202.

### DIFF
--- a/builder/proxmox/common/step_start_vm.go
+++ b/builder/proxmox/common/step_start_vm.go
@@ -287,12 +287,12 @@ func generateProxmoxDisks(disks []diskConfig) proxmox.QemuDevices {
 		if devs[idx]["type"] == "scsi" || devs[idx]["type"] == "virtio" {
 			setDeviceParamIfDefined(devs[idx], "iothread", strconv.FormatBool(disks[idx].IOThread))
 		}
-		// Proxmox API expects the value of discard to be either "off" or "on"
+		// Proxmox API expects the value of discard to be either "on" or "ignore"
 		setDeviceParamIfDefined(devs[idx], "discard", func() string {
 			if disks[idx].Discard {
 				return "on"
 			}
-			return "off"
+			return "ignore"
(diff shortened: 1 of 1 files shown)


lbajolet-hashicorp (approved):
LGTM, thanks for fixing this @jeinwag

ol3d:
When is an estimated date that these changes will be cut into a release?

lbajolet-hashicorp:
Hi @ol3d,

I believe we can release a new version of the plugin by next week.

In the meantime, if you want to experiment with the plugin, I suggest you test out the new options with one of the releases built from the latest `main`, available in https://github.com/hashicorp/packer-plugin-proxmox/actions/runs/5535351316.

Please let us know if something does not behave as documented or expected, we can probably fix it before the release of the plugin.

Thanks!

PR #204, opened 2023-07-10
Fix handling of SSD parameter.

### LINKED ISSUE #201: iothread parameter is incorrectly used as the value for the ssd parameter
#### Overview of the Issue

#199 introduced a bug where setting `io_thread = true` will also inadvertently enable ssd emulation.

#### Reproduction Steps

Build an image using this config:

```hcl
source "proxmox-iso" "debian" {
  # ...
  disks {
    io_thread    = true
    # ...
```

The resulting image will have the `ssd=1` set.

### Plugin and Packer version

Packer v1.9.1

jeinwag:
Sorry, my changes introduced another issue. This fixes it.

Closes #201.

### DIFF
--- a/builder/proxmox/common/step_start_vm.go
+++ b/builder/proxmox/common/step_start_vm.go
@@ -294,7 +294,7 @@ func generateProxmoxDisks(disks []diskConfig) proxmox.QemuDevices {
 			}
 			return "off"
 		}())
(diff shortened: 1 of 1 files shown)


lbajolet-hashicorp (approved):
Good catch, thanks for the fix

PR #205, opened 2023-07-10
Add support for PCI device pass through

romantomjak:
This PR introduces the `pci_devices` block to allow mapping host's PCI devices into the guest VM. This is often used to passthrough a graphics card or a network adapter. Devices that are mapped into a guest VM are no longer available on the host.

### DIFF
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -512,6 +512,7 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 		}
 	}
 
+	// See https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes/{node}/hardware/pci/{pciid}
 	validPCIIDre := regexp.MustCompile(`^(?:[0-9a-fA-F]{4}:)?[0-9a-fA-F]{2}:[0-9a-fA-F]{2}\.[0-9a-fA-F]$`)
 	for _, device := range c.PCIDevices {
 		if device.Host == "" && device.Mapping == "" {
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -199,13 +199,13 @@ type vgaConfig struct {
 //	```
 type pciDeviceConfig struct {
 	// The PCI ID of a host’s PCI device or a PCI virtual function. You can us the `lspci` command to list existing PCI devices. Either this or the `mapping` key must be set.
-	Host string `mapstructure:"host" required:"true"`
+	Host string `mapstructure:"host"`
 	// Override PCI device ID visible to guest.
 	DeviceID string `mapstructure:"device_id"`
 	// Pass this device in legacy IGD mode, making it the primary and exclusive graphics device in the VM. Requires `pc-i440fx` machine type and VGA set to `none`. Defaults to `false`.
 	LegacyIGD bool `mapstructure:"legacy_igd"`
 	// The ID of a cluster wide mapping. Either this or the `host` key must be set.
-	Mapping string `mapstructure:"mapping" required:"true"`
+	Mapping string `mapstructure:"mapping"`
 	// Present the device as a PCIe device (needs `q35` machine model). Defaults to `false`.
 	PCIe bool `mapstructure:"pcie"`
 	// The type of mediated device to use. An instance of this type will be created on startup of the VM and will be cleaned up when the VM stops.
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -518,6 +518,9 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 		if device.Host == "" && device.Mapping == "" {
 			errs = packersdk.MultiErrorAppend(errs, errors.New("either the host or the mapping key must be specified"))
 		}
+		if device.Host != "" && device.Mapping != "" {
+			errs = packersdk.MultiErrorAppend(errs, errors.New("the host and the mapping key cannot both be set"))
+		}
 		if device.Host != "" && !validPCIIDre.MatchString(device.Host) {
 			errs = packersdk.MultiErrorAppend(errs, errors.New("host contains invalid PCI ID"))
 		}
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -211,7 +211,7 @@ type pciDeviceConfig struct {
 	// The type of mediated device to use. An instance of this type will be created on startup of the VM and will be cleaned up when the VM stops.
 	MDEV string `mapstructure:"mdev"`
 	// Specify whether or not the device’s ROM BAR will be visible in the guest’s memory map. Defaults to `false`.
-	HideROMBAR bool `mapstructure:"rombar"`
+	HideROMBAR bool `mapstructure:"hide_rombar"`
 	// Custom PCI device rom filename (must be located in `/usr/share/kvm/`).
 	ROMFile string `mapstructure:"romfile"`
 	//Override PCI subsystem device ID visible to guest.
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: MPL-2.0
 
 //go:generate packer-sdc struct-markdown
-//go:generate packer-sdc mapstructure-to-hcl2 -type Config,NICConfig,diskConfig,rng0Config,vgaConfig,additionalISOsConfig,efiConfig
+//go:generate packer-sdc mapstructure-to-hcl2 -type Config,NICConfig,diskConfig,rng0Config,pciDeviceConfig,vgaConfig,additionalISOsConfig,efiConfig
 
 package proxmox
 
@@ -48,27 +48,28 @@ type Config struct {
 	VMName string `mapstructure:"vm_name"`
 	VMID   int    `mapstructure:"vm_id"`
 
-	Boot           string         `mapstructure:"boot"`
-	Memory         int            `mapstructure:"memory"`
-	BalloonMinimum int            `mapstructure:"ballooning_minimum"`
-	Cores          int            `mapstructure:"cores"`
-	CPUType        string         `mapstructure:"cpu_type"`
-	Sockets        int            `mapstructure:"sockets"`
-	Numa           bool           `mapstructure:"numa"`
-	OS             string         `mapstructure:"os"`
-	BIOS           string         `mapstructure:"bios"`
-	EFIConfig      efiConfig      `mapstructure:"efi_config"`
-	EFIDisk        string         `mapstructure:"efidisk"`
-	Machine        string         `mapstructure:"machine"`
-	Rng0           rng0Config     `mapstructure:"rng0"`
-	VGA            vgaConfig      `mapstructure:"vga"`
-	NICs           []NICConfig    `mapstructure:"network_adapters"`
-	Disks          []diskConfig   `mapstructure:"disks"`
-	Serials        []string       `mapstructure:"serials"`
-	Agent          config.Trilean `mapstructure:"qemu_agent"`
-	SCSIController string         `mapstructure:"scsi_controller"`
-	Onboot         bool           `mapstructure:"onboot"`
-	DisableKVM     bool           `mapstructure:"disable_kvm"`
+	Boot           string            `mapstructure:"boot"`
+	Memory         int               `mapstructure:"memory"`
+	BalloonMinimum int               `mapstructure:"ballooning_minimum"`
+	Cores          int               `mapstructure:"cores"`
+	CPUType        string            `mapstructure:"cpu_type"`
+	Sockets        int               `mapstructure:"sockets"`
+	Numa           bool              `mapstructure:"numa"`
+	OS             string            `mapstructure:"os"`
+	BIOS           string            `mapstructure:"bios"`
+	EFIConfig      efiConfig         `mapstructure:"efi_config"`
+	EFIDisk        string            `mapstructure:"efidisk"`
+	Machine        string            `mapstructure:"machine"`
+	Rng0           rng0Config        `mapstructure:"rng0"`
+	VGA            vgaConfig         `mapstructure:"vga"`
+	NICs           []NICConfig       `mapstructure:"network_adapters"`
+	Disks          []diskConfig      `mapstructure:"disks"`
+	PCIDevices     []pciDeviceConfig `mapstructure:"pci_devices"`
+	Serials        []string          `mapstructure:"serials"`
+	Agent          config.Trilean    `mapstructure:"qemu_agent"`
+	SCSIController string            `mapstructure:"scsi_controller"`
+	Onboot         bool              `mapstructure:"onboot"`
+	DisableKVM     bool              `mapstructure:"disable_kvm"`
 
 	TemplateName        string `mapstructure:"template_name"`
 	TemplateDescription string `mapstructure:"template_description"`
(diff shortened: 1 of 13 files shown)


builder/proxmox/common/step_start_vm.go
```diff
@@ -334,6 +355,13 @@ func generateProxmoxEfi(efi efiConfig) proxmox.QemuDevice {
 	return dev
 }
 
+func formatBoolPointer(v *bool) string {
+	if v == nil {
+		return ""
+	}
+	return strconv.FormatBool(*v)
+}
```
romantomjak, line 363: Didn't see any examples of optional values in the project, so came up with this. The use case is for setting a default bool value if none was given. Let me know if there is a better way to handle "unset" values 👍

docs/builders/iso.mdx
```diff
@@ -355,11 +401,7 @@ or responding to pattern `/dev/.+`. Example:
         * hdiutil (normally found in macOS)
         * oscdimg (normally found in Windows as part of the Windows ADK)
 
-
-    - `cd_content` (map[string]string) - Key/Values to add to the CD. The keys represent the paths, and the values
-      contents. It can be used alongside `cd_files`, which is useful to add large
-      files without loading them into memory. If any paths are specified by both,
-      the contents in `cd_content` will take precedence.
```
romantomjak, line 362: This is a drive-by fix for what I hope will rectify the following situation on the website:

image

builder/proxmox/common/config_test.go
```diff
@@ -338,3 +339,129 @@ func TestVMID(t *testing.T) {
...
+			expectedToFail: true,
+			pciDeviceConfig: pciDeviceConfig{
+				LegacyIGD: boolPtr(true),
+			},
+			machine: "pc",
+		},
+		{
+			expectedToFail: true,
+			pciDeviceConfig: pciDeviceConfig{
+				Host:      "0000:03:00.0",
```
lbajolet-hashicorp, line 385: Regarding the host, do we have a way to validate if they're well formed? This would be a good check to add in the `Prepare` function for the config, this way we can ensure those are not malformed by accident before attempting to build and get an obscure error from `qemu`.

builder/proxmox/common/config.go
```diff
@@ -395,8 +415,31 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
...
+		}
+	}
+
 	if errs != nil && len(errs.Errors) > 0 {
 		return nil, warnings, errs
 	}
 	return nil, warnings, nil
 }
+
+func boolPtr(v bool) *bool {
```
lbajolet-hashicorp, line 443: Since `Trilean` exists, I believe this can be removed, unless there's something that `Trilean` doesn't support and you think it's still relevant? Always open to discuss :)

docs/builders/clone.mdx
```diff
@@ -236,6 +236,51 @@ or responding to pattern `/dev/.+`. Example:
...
+    rombar        = true
+    romfile       = "vbios.bin"
+    sub_device_id = ""
+    sub_vendor_id = ""
+    vendor_id     = "15B3"
+    x_vga         = false
+  }
+  ```
+
+  - `host` (string) - The PCI ID of a host’s PCI device or a PCI virtual function. You can us the `lspci` command to list existing PCI devices. Either this or the `mapping` key must be set.
```
lbajolet-hashicorp, line 260: FYI: those options are compiled from the Go code documentation, into the partial configuration docs you've included in this PR. Those partials can be included in the documentation here, and in ISO.mdx, so you won't have to copy-paste those (which can easily diverge in the future when we modify them later).

If you want an example of this, @mabeett pushed a PR recently and did something similar in their code on PR #195: https://github.com/hashicorp/packer-plugin-proxmox/pull/195/files#diff-d395f7f43fac39fb86461a2b4239b646077ca6c2989e2fa3b0c6f174038b29ebR359

docs/builders/clone.mdx
```diff
@@ -236,6 +236,51 @@ or responding to pattern `/dev/.+`. Example:
   - `ssd` (bool) - Drive will be presented to the guest as solid-state drive
     rather than a rotational disk.
 
+- `pci_devices` (array of objects) - Passthrough a host PCI device into the VM. For example, a graphics card or a network adapter. Devices that are mapped into a guest VM are no longer available on the host. A minimal configuration only requires either the `host` or the `mapping` key to be specifed. Note: VMs with passed-through devices cannot be migrated.
```
lbajolet-hashicorp, line 239: Same for this docs exerpt, and the following example, they can be extracted into the common config documentation in the Go code, this way that'll be automatically compiled into a partial which you can include here.

lbajolet-hashicorp (changes requested):
Hi @romantomjak,

Thanks for this PR! I've taken a look at the options you're introducing and responded to your comments. I also added a few comments and suggestions that should be fixed before we can merge, but overall this looks good to me.

I'll let you address those comments, and I'll circle back later.
As always please let me know if you need any kind of help!

docs/builders/iso.mdx
lbajolet-hashicorp: FYI, PR #200 fixes this problem, so I'll merge this and you can leave this change out of yours after that!

romantomjak: Thank you, that's really useful! 🎉

builder/proxmox/common/config_test.go
romantomjak: Yeah, that's a good shout! I'll see if I can bottom it out 👍

builder/proxmox/common/step_start_vm.go
romantomjak: Thanks, these are excellent suggestions! I really like the idea of just inverting the flag 💯

builder/proxmox/common/config_test.go
romantomjak: I found a regex in proxmox API viewer that we can use 👇 
image

builder/proxmox/common/step_start_vm_test.go
```diff
@@ -441,3 +442,59 @@ func TestStartVMWithForce(t *testing.T) {
 		})
 	}
 }
+
+func TestStartVM_AssertInitialQuemuConfig(t *testing.T) {
```
romantomjak, line 446: Added this test to make sure the value is actually inverted when the config is assembled

docs/builders/clone.mdx
```diff
@@ -365,18 +368,30 @@ or responding to pattern `/dev/.+`. Example:
 
 - `machine` - (string) - Set the machine type. Supported values are 'pc' or 'q35'.
 
-#### VirtIO RNG device
```
romantomjak, line 368: De-dented header size. The config reference section has h2 headers, so I think this must be h3.

docs/builders/clone.mdx
romantomjak: Done ✅

builder/proxmox/common/config.go
```diff
@@ -171,6 +172,58 @@ type vgaConfig struct {
...
+//	```
+type pciDeviceConfig struct {
+	// The PCI ID of a host’s PCI device or a PCI virtual function. You can us the `lspci` command to list existing PCI devices. Either this or the `mapping` key must be set.
+	Host string `mapstructure:"host" required:"true"`
+	// Override PCI device ID visible to guest.
+	DeviceID string `mapstructure:"device_id"`
+	// Pass this device in legacy IGD mode, making it the primary and exclusive graphics device in the VM. Requires `pc-i440fx` machine type and VGA set to `none`. Defaults to `false`.
+	LegacyIGD bool `mapstructure:"legacy_igd"`
+	// The ID of a cluster wide mapping. Either this or the `host` key must be set.
+	Mapping string `mapstructure:"mapping" required:"true"`
```
lbajolet-hashicorp, line 208: I wonder given how the comments are phrased, are configurations with either `host` or `mapping` accepted? My understanding with `required` is that both options will be required if we note them as `required:"true"`.
In cases like these, I would suggest leaving `required` as false, and checking that one or the other is set, and not both.

lbajolet-hashicorp: I believe that your tests won't catch this since it's related to parsing configurations, this may be tested via acceptance tests, but AFAIK we don't have those setup for proxmox (we'd need some infrastructure for this).
In the meantime, could you manually check the point I bring here? I would like to make sure this won't be an issue before merging this PR.

docs/builders/iso.mdx
```diff
@@ -29,9 +29,12 @@ There are many configuration options available for the builder. They are
...
 
-In addition to the options listed here, a
-[communicator](/packer/docs/templates/legacy_json_templates/communicator) can be configured for this
-builder.
+In addition to the items listed here, you will want to look at the general
+configuration references for
+[VirtIO RNG device](#virtio-rng-device),
+[PCI Devices](#pci-devices),
+and [Communicator](/packer/docs/templates/legacy_json_templates/communicator)
+configuration references, which can be found further down the page.
```
lbajolet-hashicorp, line 37: I may misunderstand something, but this section points to the external resources, so this feels a bit out of place imho. I would suggest leaving it as-is since it points out to the communicator docs outside of the plugin, while the PCI Devices/VirtIO Rng are sections within this plugin.

builder/proxmox/common/config_test.go
```diff
@@ -421,3 +422,131 @@ func TestVMID(t *testing.T) {
...
+		{
+			pciDeviceConfig: pciDeviceConfig{
+				Host: "0000:03:00.0",
+			},
+		},
+		{
+			pciDeviceConfig: pciDeviceConfig{
+				Mapping: "someNic",
+			},
+		},
```
lbajolet-hashicorp, line 452: Would it be possible to add a test with both `Mapping` and `Host` set to make sure `Prepare` returns an error in this case?

lbajolet-hashicorp (commented):
Hi @romantomjak,

Thanks for the quick reroll, this looks good to me already :)

I left a couple more comments on the reworked PR, but as I said above, this looks really good, so I don't think there's a lot left to change here.

That being said, I'll let you have a look at what I point out/suggest, and let me know which deserve some attention.

I'll circle back later when you've had time to address those.

Thanks again for this PR!

builder/proxmox/common/config.go
romantomjak: Looks like you're right about this - [both cannot be set](https://github.com/proxmox/qemu-server/blob/742da46a44f7d882620ae0d8d8cc08ab7b3f8dcc/PVE/QemuServer/PCI.pm#L428) 👍

docs/builders/iso.mdx
romantomjak: Sorry, I don't quite follow. Are you saying to revert this change entirely or just keep the communicator in a seperate paragraph?

lbajolet-hashicorp: Oh I meant only for the communicator here, since it is a different page I think it makes sense to keep a reference to it, it's just that mixing in anchors to sections of the current page seems out-of-place to me.

romantomjak: Cool, yep that makes sense 👍

builder/proxmox/common/step_start_vm.go
```diff
@@ -295,6 +296,26 @@ func generateProxmoxDisks(disks []diskConfig) proxmox.QemuDevices {
 	return devs
 }
 
+func generateProxmoxPCIDeviceMap(devices []pciDeviceConfig) proxmox.QemuDevices {
+	devs := make(proxmox.QemuDevices)
+	for idx := range devices {
+		devs[idx] = make(proxmox.QemuDevice)
+		setDeviceParamIfDefined(devs[idx], "host", devices[idx].Host)
+		setDeviceParamIfDefined(devs[idx], "device-id", devices[idx].DeviceID)
+		setDeviceParamIfDefined(devs[idx], "legacy-igd", strconv.FormatBool(devices[idx].LegacyIGD))
```
lbajolet-hashicorp, line 305: As we discuss in #207, `setDeviceParamIfDefined` only behaves correctly when dealing with `string` values, and since these are boolean options I would suggest replacing those invocations by a `if`.

romantomjak: Sure, but I'm not entirely convinced we should do it?

I don't think it's enough to just check if the proxmox default is false and then exclude the value if the bool is false. For example, a machine has set `legacy_igd=true`. You attempt to clone it, but decide against the flag so you set it to false in packer config in which case it would be just omitted and the result would not be what you expected. To me that is worse than perhaps somewhat redundantly displaying the default value in proxmox web interface. wdyt?

builder/proxmox/common/config.go
```diff
@@ -171,6 +172,58 @@ type vgaConfig struct {
...
+// Allows passing through a host PCI device into the VM. For example, a graphics card
+// or a network adapter. Devices that are mapped into a guest VM are no longer available
+// on the host. A minimal configuration only requires either the `host` or the `mapping`
+// key to be specifed.
+//
+// Note: VMs with passed-through devices cannot be migrated.
+//
+// Example configuration (HCL):
+//
+//	```hcl
```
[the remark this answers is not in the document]
romantomjak: Sorry, that was vscode automatically inserting tabs there for some reason whenever I press ctrl+s 🤔

builder/proxmox/common/step_start_vm.go
romantomjak: Just to clarify, this is what I mean when I say "displaying default values in proxmox".

The ROM-Bar field defaults to `true`, so in proxmox when you add a PCI device it is not visible in the "hardware list view" (the line that starts with `hostpci0` behind the modal) but because we set the value to it's default when creating the VM - it gets appeneded to the config and thus also visible in the interface (even though it's at default value)

image

lbajolet-hashicorp: In this case I would say this is a matter of intent, and how we want to convey it. `setDeviceParamIfDefined` implies that we set the boolean only if it's true (the function misbehaves with such attributes, but this is an alternate discussion to have).
Your description in my opinion warrants getting rid of both the function call, and the `if` then, and we should always set their value to the boolean's (or the inverse in the case of `HideROMBAR`), and not within `setDeviceParamIfDefined`, if only to make it clear this is intended to be part of the payload, regardless if it's been set or not.
Does that make sense?

builder/proxmox/common/config.go
```diff
@@ -171,6 +172,59 @@ type vgaConfig struct {
...
+// mdev          = "some-model"
+// hide_rombar   = false
+// romfile       = "vbios.bin"
+// sub_device_id = ""
+// sub_vendor_id = ""
+// vendor_id     = "15B3"
+// x_vga         = false
+// }
+//
+//	```
```
lbajolet-hashicorp, line 200: Because of the tab here I would think this won't render as expected, I suggest this is placed on the same level as the beginning of the code block.

builder/proxmox/common/config.go
```diff
@@ -171,6 +172,59 @@ type vgaConfig struct {
...
+// on the host. A minimal configuration only requires either the `host` or the `mapping`
+// key to be specifed.
+//
+// Note: VMs with passed-through devices cannot be migrated.
+//
+// Example configuration (HCL):
+//
+// ```hcl
+// pci_devices {
+// host          = "0000:0d:00.1"
```
lbajolet-hashicorp, line 186: In HCL2 generally, the contents of a block are intented (with 2 spaces) relative to their immediate parent, I suggest we do the same here, is this OK?

romantomjak: Sorry, I can't get my editor to do that. If I add two spaces it just reindents everything extremely weirdly [like the rng0](https://github.com/romantomjak/packer-plugin-proxmox/blob/edcd09a5834cb994c848f6145d68b4573a2ade10/builder/proxmox/common/config.go#L130-L138)

builder/proxmox/common/step_start_vm.go
romantomjak: Yeah, I think so. Can you please take a look now?

builder/proxmox/common/config.go
lbajolet-hashicorp: It's OK! I can do this change in a subsequent PR either way, so I think I'll merge your PR if everything checks out, and fix this later :)

builder/proxmox/common/step_start_vm.go
lbajolet-hashicorp: Looks good to me then!

lbajolet-hashicorp (approved):
With the last changes, LGTM!

I'll merge this as soon as tests are green. Many thanks once more @romantomjak!

PR #207, opened 2023-07-15
Quick fix for issue 206

### LINKED ISSUE #206: SSD emulation not supported on virtio disk type
#### Overview of the Issue

When creating an image with disk type of virtio, option **ssd=0** is added even though it is not part of the hcl file 

#### Reproduction Steps

Setting any disk to type virtio

### Plugin and Packer version

Packer v1.9.1
Proxmox plugin 1.1.3 dev

    disks {
      disk_size         = "10G"
      format            = "raw"
      storage_pool      = "local-lvm"
      type              = "virtio"
      discard           = true
    }

### Operating system and Environment details

Ubuntu 22.04.2 LTS
Proxmox 8.0.3

Error creating VM: error creating VM: 400 Parameter verification failed., error status: {"errors":{"virtio0":"invalid format - format error\nvirtio0.ssd: property is not defined in schema and the schema does not allow additional properties\n"},"data":null} (params: map[agent:1 args: boot: cores:2 cpu:EPYC description:Packer ephemeral build VM hotplug: ide2:NAS:iso/ubuntu-22.04.2-live-server-amd64.iso,media=cdrom kvm:true machine: memory:1024 name:packer-64ad4756-f147-14e9-fba1-6ba847916e15 net0:virtio=16:C1:59:43:6D:B1,bridge=vmbr0,firewall=false numa:false onboot:false ostype:l26 scsihw:virtio-scsi-pci sockets:1 startup: tags: virtio0:local-lvm:10,format=raw,iothread=false,discard=on,ssd=0 vmid:103])

bouchardh:
This PR removes the ssd parameter when disk type is virtio as it is not supported.

Closes #206

### DIFF
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -613,11 +613,8 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 				}
 			}
 		}
-		if disk.SSD {
-			// SSD emulation is not supported on virtio device type
-			if disk.Type == "virtio" {
-				errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("SSD emulation is not supported on a virtio disk"))
-			}
+		if disk.SSD && disk.Type == "virtio" {
+			errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("SSD emulation is not supported on virtio disks"))
 		}
 		if disk.StoragePool == "" {
 			errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("disks[%d].storage_pool must be specified", idx))
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -328,6 +328,8 @@ type diskConfig struct {
 	Discard bool `mapstructure:"discard"`
 	// Drive will be presented to the guest as solid-state drive
 	// rather than a rotational disk.
+	//
+	// This cannot work with virtio disks.
 	SSD bool `mapstructure:"ssd"`
 }
 
@@ -609,6 +611,12 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 				if !(disk.Type == "scsi" || disk.Type == "virtio") {
 					errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("io thread option requires scsi or a virtio disk"))
 				}
+			}
+		}
+		if disk.SSD {
+			// SSD emulation is not supported on virtio device type
+			if disk.Type == "virtio" {
+				errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("SSD emulation is not supported on a virtio disk"))
 			}
 		}
 		if disk.StoragePool == "" {
--- a/builder/proxmox/common/step_start_vm.go
+++ b/builder/proxmox/common/step_start_vm.go
@@ -291,9 +291,9 @@ func generateProxmoxDisks(disks []diskConfig) proxmox.QemuDevices {
 			}
 			return "ignore"
 		}())
-		// SSD emulation is not supported on virtio device type
-		if devs[idx]["type"] != "virtio" {
-			setDeviceParamIfDefined(devs[idx], "ssd", strconv.FormatBool(disks[idx].SSD))
+		// Add SSD flag only if true
+		if disks[idx].SSD {
+			devs[idx]["ssd"] = "1"
 		}
 	}
 	return devs
@@ -369,6 +369,7 @@ func generateProxmoxEfi(efi efiConfig) proxmox.QemuDevice {
 }
 
 func setDeviceParamIfDefined(dev proxmox.QemuDevice, key, value string) {
+	// Empty string is considered as not defined
 	if value != "" {
 		dev[key] = value
 	}
--- a/builder/proxmox/common/step_start_vm.go
+++ b/builder/proxmox/common/step_start_vm.go
@@ -291,7 +291,10 @@ func generateProxmoxDisks(disks []diskConfig) proxmox.QemuDevices {
 			}
 			return "ignore"
 		}())
-		setDeviceParamIfDefined(devs[idx], "ssd", strconv.FormatBool(disks[idx].SSD))
+		// SSD emulation is not supported on virtio device type
+		if devs[idx]["type"] != "virtio" {
+			setDeviceParamIfDefined(devs[idx], "ssd", strconv.FormatBool(disks[idx].SSD))
+		}
 	}
 	return devs
 }
--- a/.web-docs/components/builder/clone/README.md
+++ b/.web-docs/components/builder/clone/README.md
@@ -439,6 +439,8 @@ Example:
 
 - `ssd` (bool) - Drive will be presented to the guest as solid-state drive
   rather than a rotational disk.
+  
+  This cannot work with virtio disks.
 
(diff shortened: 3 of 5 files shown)


builder/proxmox/common/step_start_vm.go
```diff
@@ -290,7 +290,10 @@ func generateProxmoxDisks(disks []diskConfig) proxmox.QemuDevices {
 			}
 			return "ignore"
 		}())
-		setDeviceParamIfDefined(devs[idx], "ssd", strconv.FormatBool(disks[idx].SSD))
+		// SSD emulation is not supported on virtio device type
+		if devs[idx]["type"] != "virtio" {
+			setDeviceParamIfDefined(devs[idx], "ssd", strconv.FormatBool(disks[idx].SSD))
+		}
```
romantomjak, line 296: Could you also please add this to [Prepare](https://github.com/hashicorp/packer-plugin-proxmox/blob/0e6bfe90fb226d6844e2dc38880dd63f95bee726/builder/proxmox/common/config.go#L266) method so that if the device is virtio and the SSD flag is set, packer fails before we attempt to start the VM?

bouchardh: I can take a look at doing it but please understand that I have no experience what so ever in golang... I made the edit because the v1.1.3 would add ssd=0 to the API call. If I am going to spend more time on this, might as well fix what I think is the real issue, setDeviceParamIfDefined is adding the option even though the option is not defined. 

I do not know if this is by design or not but all 3 boolean disk options are added as false. Are you trying to mimic the web interface defaults behaviour as this would require those option not being in the API call.

romantomjak: The goal is to prevent invalid configuration and fail early, so it's fine if the bool is present in the request with it's default value as long as it's logically allowed. Like you've already explained, the problem is that it's present in a configuration that is not supported, so we should aim to prevent that. Which is why I'm suggesting to add that to the Prepare function, so that this is caught before we attempt to start the VM. We should also update the documentation ([one](https://github.com/hashicorp/packer-plugin-proxmox/blob/0e6bfe90fb226d6844e2dc38880dd63f95bee726/docs/builders/clone.mdx#L236), [two](https://github.com/hashicorp/packer-plugin-proxmox/blob/0e6bfe90fb226d6844e2dc38880dd63f95bee726/docs/builders/iso.mdx#L247)) to reflect this restriction.

I don't think we are aiming to exclude it from the request itself to mimic what proxmox web client does. Doing that complicates the code since we'd have to update all code to use pointers or a struct that can have 3 states - unspecified, true, false and then include/exclude the values from the request. This generates additional work for little gain. Hope that's clearer now

lbajolet-hashicorp: Hey @bouchardh and @romantomjak,

The suggestion to check for SSD/Virtio in the `Prepare` method is indeed valid, if this is unsupported, we might as well catch it early and tell the user that they cannot use `ssd` in conjunction with `virtio`.

I understand this was an attempt at a quick fix, but I think we can do indeed better.

I had not taken a look at `setDeviceParamIfDefined`, but it does indeed add booleans to the request if they are formed with `strconv.FormatBool`, as we see in some other places. We should definitely fix the function (at least in terms of docs), and replace those misuses in the codebase.

In this case the problem is unrelated to `ssd` being defined, as @bouchardh points out, we can either:

1. make `setDeviceParamIfDefined` accept something other than a string
2. support values other than the empty string as `Undefined` values
3. make it clearer that `setDeviceParamIfDefined` only works with string values, and only the empty string is considered undefined. This implies replacing `setDeviceParamIfDefined` by a plain `if` when dealing with boolean values (or other types).

I think alternatives 1 or 2 would be the simplest to use, and since I don't think we have a lot of types to work with, this may be better, but I would like to not do this move hastily. I would therefore suggest we move on with 3 for now, and revisit this problem later.

@bouchardh are you OK to integrate this in your PR? I can help if needed, please let me know.

Thanks for this PR and thread, this has been instructive.

builder/proxmox/common/config.go
```diff
@@ -287,6 +287,12 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 				}
 			}
 		}
+		if disk.SSD {
+			// SSD emulation is not supported on virtio device type
+			if disk.Type == "virtio" {
```
lbajolet-hashicorp, line 292: You can make that a single directive here I would think
```suggestion
		if disk.SSD && disk.Type == "virtio" {
```

builder/proxmox/common/step_start_vm.go
bouchardh: My initial idea @lbajolet-hashicorp  was to replace all the bool variables in the diskConfig struct with Trelean to make the API call cleaner. If it's not defined, it's not there and brings attention to the actual configuration when looked at within the web ui. I understand that it does not hurt but it also defines default values as False.

I also tend to trust the developers of Proxmox to provide sensible defaults as after all, they are the experts.

Empty string for not defined makes sense to me. All we actually need is a function that would convert the Trelean to custom strings and empty could represent undefined but that would require adding a new method to the type definition (ToCustomString?) that would make it easy to provide the actual required value in the API call ie `on`/`off` or `on`/`discard` as required for the `discard` parameter.

But this is all about design.

lbajolet-hashicorp: So `Trilean` is meant for boolean attributes that don't have a sane default value in general. We generally use it for attributes that defaults to `true`, if/when that makes sense.
In cases like this one, I would think `false` is a good default value, we default on not making the disk a SSD, and therefore the intent of the `setDeviceParamIsDefined` is to not include it unless it is true, leaving the Proxmox API to figure out what to do with it.

In general, the latter makes sense, as we default to whatever the API dictates. For this case, the problem is that it always gets injected in the API because of how `setDeviceParamIsDefined`, and how it does not honour the contract of optional arguments with boolean values.

So in this particular case this is why I would think we only should inject the parameter if it is explicitely `true`, since it seems the API rejects requests with `ssd` in it if the disk type is `virtio` (even if it is false).

Regarding your suggestion to add `ToCustomString` to `Trilean`, this is trickier since `Trilean` is part of the SDK, and not this plugin. With how Go works, we cannot add arbitrary new methods to a structure outside the package it is defined in, so that would force us to write such a function in the SDK, but this doesn't necessarily make sense, as the different plugins each have their own way of describing true/false values (and inside a plugin this may change as you point out with `off`/`discard`).
So even for cases like these, I would think having a simple `if` statement around the value itself would clear any doubt here, and inject the right value in the `QemuDevice`.

builder/proxmox/common/config.go
bouchardh: Updated accordingly.

builder/proxmox/common/step_start_vm.go
bouchardh: I should read better... added comment to `setDeviceParamIfDefined` and only add the ssd parameter if true.

lbajolet-hashicorp (approved):
Hey @bouchardh,

Sorry for the long wait, I was busy on other projects.

In the current state, the code looks good to me, and addresses all the feedback we gave, so I will merge this.

There's a conflict with the docs since in the meantime we merged a PR that refactorised how it was made in the project, but that is very minor. I'll amend your commit to make it work, and merge it directly after.

Thanks for the fix, and sorry again for letting you wait so long!

PR #223, opened 2023-08-30
Run go fmt to fix comment formatting

nywilken:
Addresses the issue reported in
https://github.com/hashicorp/packer-plugin-proxmox/actions/runs/6013591915/job/16311382793?pr=222

### DIFF
--- a/builder/proxmox/clone/config.go
+++ b/builder/proxmox/clone/config.go
@@ -42,22 +42,23 @@ type Config struct {
 	Ipconfigs []cloudInitIpconfig `mapstructure:"ipconfig" required:"false"`
 }
 
-//  If you have configured more than one network interface, make sure to match the order of
-//  `network_adapters` and `ipconfig`.
+// If you have configured more than one network interface, make sure to match the order of
+// `network_adapters` and `ipconfig`.
 //
-//  Usage example (JSON):
+// Usage example (JSON):
 //
-//  ```json
-//  [
-//    {
-//      "ip": "192.168.1.55/24",
-//      "gateway": "192.168.1.1",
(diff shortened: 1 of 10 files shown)


builder/proxmox/clone/config.go
```diff
@@ -42,22 +42,23 @@ type Config struct {
...
-//      "gateway": "192.168.1.1",
-//      "ip6": "fda8:a260:6eda:20::4da/128",
-//      "gateway6": "fda8:a260:6eda:20::1"
-//    }
-//  ]
-//  ```
+// ```json
+// [
 //
+//	{
```
lbajolet-hashicorp, line 53: I hate how `gofmt` changes that nowadays, this looks so inconsistent. I wonder if there's some way we can trick it into not caring about fenced code like this

builder/proxmox/common/config.go
```diff
@@ -394,12 +394,12 @@ type rng0Config struct {
 
 // - `vga` (object) - The graphics adapter to use. Example:
```
lbajolet-hashicorp, line 395: This change looks like it could be mitigated by moving the `-` away from the docs themselves to the inclusion site. I presume `fmt` won't try to align everything based on that if we do so.

lbajolet-hashicorp: And while `fmt` does indent correctly, this borks the output.
I suggest we remove the bullet-points in the go docs. Since the parts of the docs that use this structure are already pointed at through anchors, this should be good enough.

lbajolet-hashicorp (approved):
LGTM, I'll push a change to the bullet-points as mentioned, since in the current state the fenced code gets too borked.

Other than that, we good!

PR #225, opened 2023-09-12
builder: don't use setParamIfDefined on bools

lbajolet-hashicorp:
The `setDeviceParamIfDefined' function only works on strings, and checks that the value is not empty before adding the parameter to the request.

In the builder code, in several places, there are some instances of booleans being converted to strings using `strconv.FormatBool', and then calling `setDeviceParamIfDefined' with this value, meaning that the value always gets set in the request.

This commit removes those instances in favour of a manual boolean check, which can result in adding the parameter to the request only if it is set to true.


nywilken (approved):
LGTM

PR #227, opened 2023-09-20
docs: fix builders description

KongrooParadox:
Hello folks ! :) 

This change is a proposed fix of the builders documentation.
The two descriptions in [this section](https://developer.hashicorp.com/packer/integrations/hashicorp/proxmox) were identical.

I also took the liberty of changing the description for the `iso` builder (very slightly) to be more coherent.

Let me know if the changes are OK with you ;)

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -26,11 +26,11 @@ Packer is able to target both ISO and existing Cloud-Init images.
 
 #### Builders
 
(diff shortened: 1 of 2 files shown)


lbajolet-hashicorp (approved):
Hey @KongrooParadox,

Good catch, and thanks for the contribution!

I'll merge this now

PR #232, opened 2023-09-26
Fix makefile integration docs

lbajolet-hashicorp:


lbajolet-hashicorp:
Waiting for green tests, and I merge this, release 1.0.5 of the plugin.
The release workflow was confirmed over on Hashicups, so this should fix the problem we had on our last run.

PR #238, opened 2023-11-07
Strip trailing / from proxmox url

nick-holmquist:
Accidentally adding a trailing / to the proxmox_url causes issues with additional_iso_files. A 501 can be seen and the upload fails.
Unfortunately, that error is a red herring. While the url is logged in debug mode, it is simply a log print and not obvious as it doesn't actually point out the error 

Can replicate by simply specifying a / at the end of the proxmox_url config
proxmox_url ="https://my-proxmox.my-domain:8006/api2/json/"

The upload request becomes; eg. POST /api2/json//nodes/pve/storage/local/upload
Note the json// with the extra slash.

Stripping the trailing slash should resolve this issue when someone accidentally copies/pastes the url to the config.

Edit: Another option would be to check the string format in general and give an error to resolve before continuing.

### DIFF
--- a/builder/proxmox/common/client.go
+++ b/builder/proxmox/common/client.go
@@ -6,6 +6,7 @@ package proxmox
 import (
 	"crypto/tls"
 	"log"
+	"strings"
 
 	"github.com/Telmate/proxmox-api-go/proxmox"
 )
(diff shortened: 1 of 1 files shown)


nick-holmquist:
> Good catch @nick-holmquist!
> 
> The fix looks good to me, I'll wait for tests to go green, and merge this after.
> 
> I'll release a new version of the plugin today since this seems like a fix to an easy-to-hit problem.
> 
> Thanks for submitting this PR!

The other option would be to sanitize/verify the url on submission (regex match) and actually error out properly. I don't know GoLang to be able to do that proficiently though.   I spent enough time trying to learn/debug Go on the fly just to find I had an extra slash in my url! 

Thanks for the quick review!

PR #245, opened 2023-12-31
Add 'args' as optional configuration parameter when creating a VM

### LINKED ISSUE #243: Can I set "args=" to specify additional arguments in the Proxmox plugin
(Also posted this on Google Groups - https://groups.google.com/g/packer-tool/c/-8a52KYfkXA - but that seems to be filled with spam...)

Hello,

I am trying to setup deploying a macOS VM using Proxmox (legally, it is running on an old iMac, which fits Apple's license). For this I am following this guide:
https://www.youtube.com/watch?v=68R2SdbFj-8&ab_channel=i12bretro

I was able to run this manually, but since we will reinstalling the macOS instances regularly, I would like to deploy this using Packer.

One of the steps in the guide is to add an additional line in the VM conf file that has 
  args: ...
however, I do not see any means to pass this from Packer.
I looked in the Proxmox API and it seems to be a valid part of the VM config.

Did I miss something, or is this simply not supported by the Proxmox plugin?

Thanks,

Ronald

rhpijnacker:
To create macOS VMs in Proxmox, the `args` parameter needs to be set to pass some additional values to the underlying KVM. This PR adds support for setting this in the packer source.

This builds on https://github.com/Telmate/proxmox-api-go/pull/117/files, which already makes it possible to pass `Args` in the `ConfigQemu` struct.

Closes #243

### DIFF
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -97,6 +97,10 @@ type Config struct {
 	// `debian-12;template`.
 	Tags string `mapstructure:"tags"`
 
+	// Arbitrary arguments passed to KVM. For example
+	// ` -no-reboot -smbios type=0,vendor=FOO`.
+	// 	Note: this option is for experts only.
+	Args string `mapstructure:"args"`
 	// Override default boot order. Format example `order=virtio0;ide2;net0`.
 	// Prior to Proxmox 6.2-15 the format was `cdn` (c:CDROM -> d:Disk -> n:Network)
 	Boot string `mapstructure:"boot"`
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -98,7 +98,7 @@ type Config struct {
 	Tags string `mapstructure:"tags"`
 
 	// Arbitrary arguments passed to KVM. For example
-	// ` -no-reboot -smbios type=0,vendor=FOO`.
+	// `-no-reboot -smbios type=0,vendor=FOO`.
 	// 	Note: this option is for experts only.
 	Args string `mapstructure:"args"`
 	// Override default boot order. Format example `order=virtio0;ide2;net0`.
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -97,10 +97,6 @@ type Config struct {
 	// `debian-12;template`.
 	Tags string `mapstructure:"tags"`
 
-	// Arbitrary arguments passed to KVM. For example
-	// `-no-reboot -smbios type=0,vendor=FOO`.
-	// 	Note: this option is for experts only.
-	Args string `mapstructure:"args"`
 	// Override default boot order. Format example `order=virtio0;ide2;net0`.
 	// Prior to Proxmox 6.2-15 the format was `cdn` (c:CDROM -> d:Disk -> n:Network)
 	Boot string `mapstructure:"boot"`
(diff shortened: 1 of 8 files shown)


rhpijnacker:
I tested this by deploying a VM that includes the `args=` property.
I'm not sure what kind of unit test to add.

builder/proxmox/common/config.go
```diff
@@ -97,6 +97,10 @@ type Config struct {
 	// `debian-12;template`.
 	Tags string `mapstructure:"tags"`
 
+	// Arbitrary arguments passed to KVM. For example
+	// ` -no-reboot -smbios type=0,vendor=FOO`.
```
lbajolet-hashicorp, line 101: Out of curiosity, is the space in front of `no-reboot` voluntary or is it a typo? If it is intended, I would suggest documenting it clearly so users aren't surprised in case they don't add one and it fails unexpectedly?

builder/proxmox/common/config.go
```diff
@@ -97,6 +97,10 @@ type Config struct {
...
+	// 	Note: this option is for experts only.
+	Args string `mapstructure:"args"`
```
lbajolet-hashicorp, line 103: Bit nitpicky on my part, but just for clarity, since this is for extra raw args, and since this is for advanced use-cases only, I would suggest making the name more explicit here, something along the lines of `qemu_extra_args`.

lbajolet-hashicorp (commented):
Hi @rhpijnacker,

Thanks for the PR! I left a few comments which err more on the nitpick side of things than anything else, the code itself looks good to me.
~To be sure, have you tested those changes locally? Does it work as you would expect?~ Nevermind this, I just saw your second message, all clear here, I don't think there's a lot of tests for this plugin already (might be good to have some eventually) so I would think we'll rely on good will and manual testing for now.

builder/proxmox/common/config.go
rhpijnacker-philips: There is no reason for that space. Removed it.

builder/proxmox/common/config.go
rhpijnacker-philips: Sure. Renamed it to `qemu_additional_args` (to stay consistent with e.g. `additional_iso_files`).
Renamed the internal structure-property to `AdditionalArgs`.

rhpijnacker-philips: FYI: Succesfully tested it manually with the new name.

rhpijnacker: Sure. Renamed it to `qemu_additional_args` (to stay consistent with e.g. `additional_iso_files`).
Renamed the internal structure-property to `AdditionalArgs`.

rhpijnacker: FYI: Succesfully tested it manually with the new name.

builder/proxmox/common/config.go
rhpijnacker: There is no reason for that space. Removed it.

lbajolet-hashicorp (approved):
Thanks for the reroll @rhpijnacker!

This looks good to me, I'll merge this PR now.

lbajolet-hashicorp:
FYI: CI was complaining that the generated code differed, I presume you had amended the docs and hcl2spec files manually?
If you want to contribute to the plugin later, I suggest using the `make generate` target for this, we're using a tool from our SDK to generate the docs partials and the hcl2spec files from the Go code so we don't need to manually maintain those.

I've pushed the change on your branch, on top of your commits, so we can merge this now.

PR #251, opened 2024-02-01
add the before_start_hook

### LINKED ISSUE #198: Give us enough rope to edit the raw VM creation request
#### Community Note

#### Description

Sometimes we need to modify arguments that are not supported by the plugin. For such times, can there be a way to modify the https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes/{node}/qemu request before submitting it to the proxmox api?

#### Use Case(s)

This is required to be able to fiddle with the proxmox `args` conf and to arbitrarily patch other settings.

#### Potential configuration

```
# this would would execute the script, passing json into stdin and would expect json in stdout.
# the json out is the one actually sent to proxmox.
configuration_transformation_script = ["./patch-vm-conf.sh"]
```

rgl:
this is a PoC for https://github.com/hashicorp/packer-plugin-proxmox/issues/198.

I'm looking for suggestions about the name of the configuration property, currently, its called `before_start_hook`.

if you agree with this change, I'll later will be adding the missing documentation.

the `before_start_hook` configuration setting contains the command that will be executed after the vm is configured, but before its started.

the command receives the current vm configuration as a json in its stdin

the hook command is expected to return a json with the modified configuration, typically, this document should only contain the properties that you want to actually modify.

the following example is modifying the `ide0` and `ide1` media from `cdrom` to `disk`, something that is useful when we have a disk image instead of an iso:

```hcl
source "proxmox-iso" "macos-amd64" {
  # ...
  before_start_hook = [
    "python3",
    "-c",
    <<-EOF
    import json, sys
    config = json.load(sys.stdin)
    print(json.dumps({
      "ide0": config["ide0"].replace("media=cdrom", "media=disk"),
      "ide2": config["ide2"].replace("media=cdrom", "media=disk"),
    }))
    EOF
  ]
```

Closes #198


rhpijnacker-philips:
It seems to me this is a bit of a kludge.
It means that we are not currently able to express the desired configuration with the existing parameters.

rgl:
@rhpijnacker-philips yes, like described at https://github.com/hashicorp/packer-plugin-proxmox/issues/198, this is an escape hatch, to use in those cases where there is no other way.

medioxor:
i am desperate for this! very excited as i'd like to automate mac deployment

PR #253, opened 2024-03-08
.go-version: bump to 1.21.8

lbajolet-hashicorp:


lbajolet-hashicorp:
Managed externally, no need to update it via a PR.

PR #256, opened 2024-03-24
[Issue 74] Create var for Cloud-Init Disk Type

### LINKED ISSUE #74: Use SCSI controller for Cloud-Init
#### Community Note

#### Description

Cloud-init drives should be attached to a SCSI controller instead of an IDE controller.

#### Use Case(s)

RHEL 8 doesn't support (some?) IDE drives anymore, causing Cloud-Init to abort while starting, since the Cloud-init drive is missing.

#### Potential configuration

\-

#### Potential References

https://bugs.centos.org/view.php?id=16898

xrayj11:
This change will allow users to specify a `cloud_init_disk_type`, which determines which type (ide, sata, scsi) of drive is created for Cloud-Init. If no value is given or the variable is not provided, the Cloud-Init drive will be created on the first available ide drive, which was how it operated before.
This change should not break any existing configurations that use Cloud-Init, and only provides additional customization to specify a different drive type than the default `ide`

These changes passed the `make test`, as well as packer validation after installing the newly packaged plugin and declaring the `cloud_init_disk_type` var in my hcl file.
I have tested this change with the proxmox-iso builder on each disk type (ide, sata, scsi) as well as not providing the var to ensure it will not break existing configurations. Each option resulted in the expected outcome, where the Cloud-Init drive was created with the type specified by the var.

Closes #74 - By giving the option of SCSI for Cloud-Init

### DIFF
--- a/.web-docs/components/builder/iso/README.md
+++ b/.web-docs/components/builder/iso/README.md
@@ -198,7 +198,7 @@ in the image's Cloud-Init settings for provisioning.
   to store the Cloud-Init CDROM on. If not given, the storage pool of the boot device will be used.
 
 - `cloud_init_disk_type` (string) - The type of Cloud-Init disk. Can be `scsi`, `sata`, or `ide`
-  If not given, defaults to `ide`.
+  Defaults to `ide`.
 
 - `additional_iso_files` ([]additionalISOsConfig) - Additional ISO files attached to the virtual machine.
   See [Additional ISO Files](#additional-iso-files).
--- a/.web-docs/components/builder/iso/README.md
+++ b/.web-docs/components/builder/iso/README.md
@@ -197,6 +197,9 @@ in the image's Cloud-Init settings for provisioning.
 - `cloud_init_storage_pool` (string) - Name of the Proxmox storage pool
   to store the Cloud-Init CDROM on. If not given, the storage pool of the boot device will be used.
 
+- `cloud_init_disk_type` (string) - The type of Cloud-Init disk.
+  Can be `scsi`, `sata`, or `ide`. If not given, defaults to `ide`.
+
 - `additional_iso_files` ([]additionalISOsConfig) - Additional ISO files attached to the virtual machine.
   See [Additional ISO Files](#additional-iso-files).
 
--- a/.web-docs/components/builder/iso/README.md
+++ b/.web-docs/components/builder/iso/README.md
@@ -197,8 +197,8 @@ in the image's Cloud-Init settings for provisioning.
 - `cloud_init_storage_pool` (string) - Name of the Proxmox storage pool
   to store the Cloud-Init CDROM on. If not given, the storage pool of the boot device will be used.
 
-- `cloud_init_disk_type` (string) - The type of Cloud-Init disk.
-  Can be `scsi`, `sata`, or `ide`. If not given, defaults to `ide`.
+- `cloud_init_disk_type` (string) - The type of Cloud-Init disk. Can be `scsi`, `sata`, or `ide`
+  If not given, defaults to `ide`.
 
 - `additional_iso_files` ([]additionalISOsConfig) - Additional ISO files attached to the virtual machine.
   See [Additional ISO Files](#additional-iso-files).
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -654,12 +654,12 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 	}
 	if c.CloudInit {
 		switch c.CloudInitDiskType {
-			case "ide", "scsi", "sata":
-			case "":
-				log.Printf("Cloud-Init disk type not set, using default 'ide'")
-				c.CloudInitDiskType = "ide"
-			default:
-				errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("invalid value for `cloud_init_disk_type` %q: only one of 'ide', 'scsi', 'sata' is valid", c.CloudInitDiskType))
+		case "ide", "scsi", "sata":
+		case "":
+			log.Printf("Cloud-Init disk type not set, using default 'ide'")
+			c.CloudInitDiskType = "ide"
+		default:
+			errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("invalid value for `cloud_init_disk_type` %q: only one of 'ide', 'scsi', 'sata' is valid", c.CloudInitDiskType))
 		}
 	}
 
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -652,14 +652,14 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 		log.Printf("SCSI controller not set, using default 'lsi'")
 		c.SCSIController = "lsi"
 	}
-	if c.CloudInit == true {
-		if c.CloudInitDiskType == "" {
-			log.Printf("Cloud-Init disk type not set, using default 'ide'")
-			c.CloudInitDiskType = "ide"
-		} else {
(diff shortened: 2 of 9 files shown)


builder/proxmox/common/step_finalize_template_config.go
```diff
@@ -60,10 +60,27 @@ func (s *stepFinalizeTemplateConfig) Run(ctx context.Context, state multistep.St
...
+					sataController := fmt.Sprintf("sata%d", i)
+					diskControllers = append(diskControllers, sataController)
+				}
+			// and up to 31 SCSI controllers (0 - 30)
+			case "scsi":
+				for i := 0; i < 31; i++ {
+					scsiController := fmt.Sprintf("scsi%d", i)
+					diskControllers = append(diskControllers, scsiController)
+				}
+			default:
```
lbajolet-hashicorp, line 77: I would suggest defaulting to ide only if it's empty or explicitly `ide`, otherwise this allows for anything and I can see this being a surprise potentially.

```suggestion
			"", "ide":
```

lbajolet-hashicorp: EDIT: nevermind I notice we're in `Run` here, I'd suggest then having that check/error in the `Prepare` function then, so we can set `ide` if unset for example, and we can error if the option has an unsupported value.

builder/proxmox/common/config.go
```diff
@@ -183,6 +183,9 @@ type Config struct {
 	// Name of the Proxmox storage pool
 	// to store the Cloud-Init CDROM on. If not given, the storage pool of the boot device will be used.
 	CloudInitStoragePool string `mapstructure:"cloud_init_storage_pool"`
+	// The type of Cloud-Init disk. Can be `scsi`, `sata`, or `ide`
+	// If not given, defaults to `ide`.
```
lbajolet-hashicorp, line 187: Nit to make it more concise.

```suggestion
	// Defaults to `ide`.
```

lbajolet-hashicorp (commented):
Hi @xrayj11,

Thanks for the PR, and sorry I took a long time to review it.

This looks good to me overall, I have left a couple comments/suggestions, feel free to address them, and I can do another review when you've had time to respond.

Thanks again!

builder/proxmox/common/step_finalize_template_config.go
xrayj11: I have added a default, check, and error in the `Prepare` function. Does this look good?
I tested it with a `CloudInitDiskType` value that was not supported (not `scsi`,`sata`, or `ide`) and it stopped packer with the error message I added. I also confirmed that blanks still default to `ide`.
https://github.com/hashicorp/packer-plugin-proxmox/pull/256/commits/e03831f1da79eed6ff899fff8d4ad9258d0087ff

builder/proxmox/common/config.go
```diff
@@ -652,6 +652,16 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 		log.Printf("SCSI controller not set, using default 'lsi'")
 		c.SCSIController = "lsi"
 	}
+	if c.CloudInit == true {
```
lbajolet-hashicorp, line 655:
```suggestion
	if c.CloudInit {
```

builder/proxmox/common/config.go
```diff
@@ -652,6 +652,16 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
...
+		if c.CloudInitDiskType == "" {
```
lbajolet-hashicorp, line 656: I believe the whole `if c.CloudInit` block can be simplified with a `switch`?

```go
switch c.CloudInitDiskType {
	case "ide", "scsi", "sata":
	case "":
		c.CloudInitDiskType = "ide"
	default:
		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("invalid value for `cloud_init_disk_type` %q: only one of 'ide', 'scsi', and 'sata' is valid", c.CloudInitDiskType))
}
```

builder/proxmox/common/step_finalize_template_config.go
lbajolet-hashicorp: I would still suggest expliciting `ide` for this case, just in case something else manages to go through the `Prepare`, so that it doesn't go into `ide` mode when reaching this step.
Adding a `default` with an error is probably for the best for later developments, as this will error in case we update `Prepare` but forget to do the same here.
```suggestion
			"ide":
				// Unspecified disk type defaults to "ide"
				diskControllers = []string{"ide0", "ide1", "ide2", "ide3"}
			default:
				state.Put("error", fmt.Errorf("unsupported disk type %q", c.CloudInitDiskType))
				return multistep.ActionHalt
```

.web-docs/components/builder/iso/README.md
```diff
@@ -197,6 +197,9 @@ in the image's Cloud-Init settings for provisioning.
 - `cloud_init_storage_pool` (string) - Name of the Proxmox storage pool
   to store the Cloud-Init CDROM on. If not given, the storage pool of the boot device will be used.
 
+- `cloud_init_disk_type` (string) - The type of Cloud-Init disk. Can be `scsi`, `sata`, or `ide`
+  If not given, defaults to `ide`.
```
lbajolet-hashicorp, line 201: I believe this should update itself based on the comment from `config.go`, have you run `make generate` to update that?

lbajolet-hashicorp (commented):
Hi @xrayj11,

Thanks for the update, I left a few more comments on the code for simplification and hardening purposes, please feel free to take a look and address them when you can.

I'll do another review when you've had time to do this.

Thanks!

lbajolet-hashicorp:
Hi @xrayj11,

The rework looks good to me, however it seems there are failures during test, also the file should be linted/goimported so the CI is happy, may I ask you to fix those before we can merge?

xrayj11:
Hey @lbajolet-hashicorp, I've been trying to figure out the issue with the failed `all_options_with_cloud-init` test, and I think I've narrowed it down. Although we have a case to default to `ide` when `CloudInitDiskType` is blank in the Prepare function of config.go, the test is not using it. This means it's going into the Run function with a blank/undefined value and triggering the default case, and erroring out.
When I define a valid `CloudInitDiskType` here the test passes:
https://github.com/hashicorp/packer-plugin-proxmox/blob/main/builder/proxmox/common/step_finalize_template_config_test.go#L106-L112
Could we do that, or is there another way we should be testing this since the value can be changed/defaulted by config.go?
Also, I've fixed the formatting/goimport issues, they'll be added once we figure out the test.

lbajolet-hashicorp:
Ah I see @xrayj11, in this case I would suggest either of the two alternatives:

1. Add a call to `Prepare` before calling `Run` here: this way only valid configs will reach `Run`. This is more in line with what we expect from normal plugin usage, as such it cannot hurt imho. This could conflict with other tests though, so I would advise not doing this if this becomes a problem for other test cases.
2. Add a `CloudInitDiskType` to the config, so it is not empty when executing `Run` in this case. Given this option is supposed to be vetted/set by `Prepare`, if the aim of the test is only to test `Run` in a valid case, this may be a valid change.

I'll let you pick which approach seems more appropriate to you, feel free to ping me when you've settled on either!

xrayj11:
I looked into trying to call `Prepare` in the test(s), but it looks like it's going to require a bit of a rewrite of how all the tests are done. I do agree it would be a better implementation of testing since it is more in line with how it actually runs. Since that's out of the scope of this issue, I've just defined `CloudInitDiskType: "ide"` on any test that uses `CloudInit: true`.
All tests ran good on my side, and in actual packer runs in my environment the plugin still works without defining a `CloudInitDiskType`, so existing configs will not be affected. Let me know if you see anything else that needs changed, or if this is can be merged.
Thanks for the help @lbajolet-hashicorp

lbajolet-hashicorp (approved):

Issue #187, opened 2023-04-14
CD/DVD drive assign scsi instead of ide


oscar-vlugt:
I am trying to use packer on a proxmox cluster running on Raspberry PIs which don't have any IDE support. Normally I would just manually create a virtual machine and then remove the cd/dvd drive and assign SCSI to circumvent the following error:

`kvm: -device ide-cd,bus=ide.1,unit=0,drive=drive-ide2,id=ide2,bootindex=101: Bus 'ide.1' not found
`

Huge fan of Packer, and want to roll it out for my ARM images. Is there any possibility for me to add to my pkr.hcl file to set CD/DVD upon creation to use the Bus SCSI instead of IDE? 

My packer file contains : 

```source "proxmox-iso" "debian"{
    # Proxmox Connection Settings
    proxmox_url              = "${var.proxmox_api_url}"
    username                 = "${var.proxmox_api_token_id}"
    token                    = "${var.proxmox_api_token_secret}"
    insecure_skip_tls_verify = true
    # Packer VM Configuration
    bios = "ovmf"
    boot_command      = [
        "<esc> auto preseed/url=http://{{ .HTTPIP }}:{{ .HTTPPort }}/Debian-11.x/preseed.cfg<enter>"
    ]
    boot_wait                = "3s"
    cores                    = 2
    cpu_type                 = "host"
    disks {
        disk_size    = "4G"
        format       = "qcow2"
        storage_pool = "local"
        type         = "virtio"
    }
    http_directory           = "http"
    http_port_min            = "8000"
    http_port_max            = "8000"
    http_bind_address        = "0.0.0.0"
    iso_file                 = "local:iso/debian-11.6.0-arm64-netinst.iso"
    memory                   = 2048
    network_adapters {
        bridge = "vmbr0"
        model  = "virtio"
    }
    node                     = "pve002"
    onboot                   = true
    qemu_agent               = true
    scsi_controller          = "virtio-scsi-pci"
    sockets                  = 1
    ssh_username             = "root"
    ssh_password             = "*"
    ssh_timeout              = "3600s"
    template_description     = "Debian 11.x, generated on ${timestamp()}"
    template_name            = "TMPL-Debian11.x"
    unmount_iso              = true
    vm_name                  = "TMPL-Debian11.x"
    vm_id                    = 9980
}

build {
    sources = [
        "source.proxmox-iso.debian"
    ]
}

sebastian-de:
[...]

Issue #80, opened 2022-04-08
Add feature to remove cd-rom that was used for unmount_iso


npdev453:
unmount_iso did not remove cd-rom that whas used for building from iso
but in sometimes template should not have any cd-roms

mpywell:
I think this is especially needed for additional_iso_files blocks, current builds with one or more additional_iso_files blocks end up with at least two (boot ISO and one or more additional_iso_files) cdrom devices attached to the template output. 

Thoughts on making the behaviour of unmount = true for additional_iso_files blocks also remove the cdrom device? For the boot ISO device (ide2) I could see the value in keeping one cdrom device attached to a template by default, maybe removing ide2 could be an opt-in setting.

PR #267, opened 2024-05-31
[Issues 257, 263, 187, 80] - AsyncIO disk config support, ISO Storage uplift

### LINKED ISSUE #257: Add aio disk configuration available
#### Description

To have a proper golden image we need to set **aio = "native"** which is not the default configuration in Proxmox. Please add this feature

#### Use Case(s)

**aio = native** is way faster than default Proxmox **io_uring** configuration

### LINKED ISSUE #263: Aarch64 (ARM64) unavailable due to lack of IDE drivers for the architecture. CDROM driver should be set via scsi or virtio.
When filing a bug, please include the following headings if possible. Any
example text in this template can be deleted.

#### Overview of the Issue

When attempting to create a virtual machine using aarch64 as the architecture, the packer script fails:

```
kvm: -device ide-cd,bus=ide.1,unit=0,drive=drive-ide2,id=ide2: Bus 'ide.1' not found
TASK ERROR: start failed: QEMU exited with code 1
```

The same build process works under x86_64. These builds are using the UEFI firmware, not BIOS. It appears for aarch64, there needs to be an option to set the bus for the that device in order to accommodate  native, non-x86_64 architectures.  This is not using aarch64 emulation under x86_64.

#### Reproduction Steps

Run packer script on an aarch64 machine (not emulated).  The current build is meant to run packer to provision a template and use a kickstart script to install and configure a Rocky Linux 9.3 VM/template.

### Plugin and Packer version

- Packer v1.10.3
-  Proxmox plugin v1.1.7 (packer-plugin-proxmox_v1.1.7_x5.0_darwin_amd64)

```
source "proxmox-iso" "rocky9-kickstart" {
  <enter><wait><enter>"]
  # For UEFI
  boot_command = [
    "<up>",
    "e",
    "<down><down><end><wait>",
    "text inst.ks=http://{{ .HTTPIP }}:{{ .HTTPPort }}/inst.ks",
    "<enter><wait><leftCtrlOn>x<leftCtrlOff>"
  ]
  boot_wait    = "3s"
  cores        = "2"
  cpu_type     = "host"
  bios         = "ovmf"
  boot         = "order=virtio0;virtio1"
  disks {
    disk_size         = "128G"
    format            = "${var.proxmox_storage_format}"
    storage_pool      = "${var.proxmox_storage_pool}"
    type              = "virtio"
  }
  efi_config {
    efi_storage_pool        = "local"
    efi_type                      = "4m"
    pre_enrolled_keys      = false
  } 
  http_directory           = "packer/rocky9-aarch64"
  insecure_skip_tls_verify = true
  iso_file                 = "${var.proxmox_iso_pool}/${var.rocky_image}"
  memory                   = "2048"
  network_adapters {
    bridge = "vmbr0"
    model  = "virtio"
  }
  node                 = "${var.proxmox_node}"
  os                   = "l26"
  password             = "${var.proxmox_password}"
  proxmox_url          = "${var.proxmox_url}"
  qemu_agent           = true
  scsi_controller      = "virtio-scsi-pci"
  ssh_password         = "Packer"
  ssh_port             = 22
  ssh_timeout          = "30m"
  ssh_username         = "root"
  template_description = "${var.template_description}"
  template_name        = "${var.template_name}"
  unmount_iso          = true
  cloud_init           = true
  cloud_init_storage_pool = "${var.proxmox_storage_pool}"
  username             = "${var.proxmox_username}"
}

build {
  sources = ["source.proxmox-iso.rocky9-kickstart"]
  name = "proxmox-rl9-aarch64"
  provisioner "shell" {
    inline = ["yum install -y cloud-init qemu-guest-agent cloud-utils-growpart gdisk", "systemctl enable qemu-guest-agent", "shred -u /etc/ssh/*_key /etc/ssh/*_key.pub", "rm -f /var/run/utmp", ">/var/log/lastlog", ">/var/log/wtmp", ">/var/log/btmp", "rm -rf /tmp/* /var/tmp/*", "unset HISTFILE; rm -rf /home/*/.*history /root/.*history", "rm -f /root/*ks", "passwd -d root", "passwd -l root", "rm -f /etc/ssh/ssh_config.d/allow-root-ssh.conf"]
  }
}
```

To reproduced the error
```bash
packer init packer/rocky9-aarch64/packer.pkr.hcl
packer build -var-file variables-aarch64.pkrvars.hcl packer/rocky9-aarch64/packer.pkr.hcl
```
In the shell console, this is the result:
```
proxmox-rl9-aarch64.proxmox-iso.rocky9-kickstart: output will be in this color.

==> proxmox-rl9-aarch64.proxmox-iso.rocky9-kickstart: Creating VM
==> proxmox-rl9-aarch64.proxmox-iso.rocky9-kickstart: No VM ID given, getting next free from Proxmox
==> proxmox-rl9-aarch64.proxmox-iso.rocky9-kickstart: Starting VM
==> proxmox-rl9-aarch64.proxmox-iso.rocky9-kickstart: Error starting VM: start failed: QEMU exited with code 1
==> proxmox-rl9-aarch64.proxmox-iso.rocky9-kickstart: Stopping VM
==> proxmox-rl9-aarch64.proxmox-iso.rocky9-kickstart: Deleting VM
Build 'proxmox-rl9-aarch64.proxmox-iso.rocky9-kickstart' errored after 24 seconds 741 milliseconds: Error starting VM: start failed: QEMU exited with code 1
```

In the proxmox console, the build fails, and shows that the CDROM is trying to be mounted on an ide bus:

```
kvm: -device ide-cd,bus=ide.1,unit=0,drive=drive-ide2,id=ide2: Bus 'ide.1' not found
TASK ERROR: start failed: QEMU exited with code 1
```

### Operating system and Environment details

**CPU(s):** Cortex-A72 (also tested under ARMv8 (M128-26))
**System:** SolidRun CEX7 Platform (Also tested on Ampere Altra)
**Kernel Version:** 6.8.9-1.el9.elrepo.aarch64 and  Linux 6.1.0-20-arm64 (2024-04-11)
**Proxmox version:** pve-manager/8.1.7/ee1c3736ef6a6541

Relevant log information:
```
proxmox-rl9-aarch64.proxmox-iso.rocky9-kickstart: output will be in this color.

2024/05/09 10:17:52 Build debug mode: false
2024/05/09 10:17:52 Force build: false
2024/05/09 10:17:52 On error:
2024/05/09 10:17:52 Waiting on builds to complete...
2024/05/09 10:17:52 Starting build run: proxmox-rl9-aarch64.proxmox-iso.rocky9-kickstart
2024/05/09 10:17:52 Running builder: proxmox-iso
2024/05/09 10:17:52 [INFO] (telemetry) Starting builder proxmox-iso.rocky9-kickstart
2024/05/09 10:17:52 packer-plugin-proxmox_v1.1.7_x5.0_darwin_amd64 plugin: 2024/05/09 10:17:52 using password auth
2024/05/09 10:17:52 packer-plugin-proxmox_v1.1.7_x5.0_darwin_amd64 plugin: 2024/05/09 10:17:52 No URLs were provided to Step Download. Continuing...
==> proxmox-rl9-aarch64.proxmox-iso.rocky9-kickstart: Creating VM
==> proxmox-rl9-aarch64.proxmox-iso.rocky9-kickstart: No VM ID given, getting next free from Proxmox
==> proxmox-rl9-aarch64.proxmox-iso.rocky9-kickstart: Starting VM
==> proxmox-rl9-aarch64.proxmox-iso.rocky9-kickstart: Error starting VM: start failed: QEMU exited with code 1
==> proxmox-rl9-aarch64.proxmox-iso.rocky9-kickstart: Stopping VM
==> proxmox-rl9-aarch64.proxmox-iso.rocky9-kickstart: Deleting VM
2024/05/09 10:18:15 [INFO] (telemetry) ending proxmox-iso.rocky9-kickstart
Build 'proxmox-rl9-aarch64.proxmox-iso.rocky9-kickstart' errored after 22 seconds 753 milliseconds: Error starting VM: start failed: QEMU exited with code 1
==> Wait completed after 22 seconds 754 milliseconds

2024/05/09 10:18:15 machine readable: error-count []string{"1"}
==> Some builds didn't complete successfully and had errors:
==> Wait completed after 22 seconds 754 milliseconds

2024/05/09 10:18:15 machine readable: proxmox-rl9-aarch64.proxmox-iso.rocky9-kickstart,error []string{"Error starting VM: start failed: QEMU exited with code 1"}
==> Some builds didn't complete successfully and had errors:
==> Builds finished but no artifacts were created.
--> proxmox-rl9-aarch64.proxmox-iso.rocky9-kickstart: Error starting VM: start failed: QEMU exited with code 1

==> Builds finished but no artifacts were created.
2024/05/09 10:18:15 [INFO] (telemetry) Finalizing.
2024/05/09 10:18:15 waiting for all plugin processes to complete...
2024/05/09 10:18:15 /usr/local/bin/packer: plugin process exited
```

mpywell:
Hey @lbajolet-hashicorp!

As mentioned in #260, here is my follow up to close out some related storage open issues.

To resolve #257 I've added AsyncIO as an option to the storage uplift from #260.

To resolve #263 and #187 I've added a config option to the proxmox-iso builder's ISO device to specify a device type and bus index in the same style as the Additional ISO Files block `device` attribute.

To address #80 I've changed the default behaviour of the `unmount` option for ISOs to remove the empty CDROM device. A new config option `unmount_keep_device` is available if there's a desire to keep empty CDROM devices in the produced VM template.

The commits for #263, #187 and #80 have a description that further explains their changes.

Closes #257, #263, #187, #80


builder/proxmox/common/config.go
```diff
@@ -682,28 +683,6 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
...
-		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("maximum 2 IDE disks supported (ide2,3 reserved for ISOs)"))
-	}
-	if sataCount > 6 {
-		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("maximum 6 SATA disks supported"))
-	}
-	if scsiCount > 31 {
-		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("maximum 31 SCSI disks supported"))
-	}
-	if virtIOCount > 16 {
-		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("maximum 16 VirtIO disks supported"))
```
lbajolet-hashicorp, line 697: Is there a reason to remove those checks? I suspect people may not hit the limit too easily, but it might be a good idea to error at this step so it doesn't error later during the build process (no cleanup needed and it happens before the build starts if it errors during `Prepare`

builder/proxmox/common/step_start_vm.go
```diff
@@ -280,17 +265,73 @@ func generateProxmoxNetworkAdapters(nics []NICConfig) proxmox.QemuDevices {
...
+			rb := regexp.MustCompile(`\d+`)
+			index, _ := strconv.Atoi(rb.FindString(iso.Device))
+
+			log.Printf("Mapping Additional ISO to %s%d", device, index)
+			switch device {
+			case "ide":
+				dev := proxmox.QemuIdeStorage{
+					CdRom: bootIso,
+				}
+				reflect.
```
lbajolet-hashicorp, line 306: I wonder given this is very similar to what you added in an earlier PR, could we factorise this code that adds the device to the VM spec?

builder/proxmox/iso/config.go
```diff
@@ -23,6 +27,12 @@ type Config struct {
 	// `local:iso/Fedora-Server-dvd-x86_64-29-1.2.iso`.
 	// Either `iso_file` OR `iso_url` must be specifed.
 	ISOFile string `mapstructure:"iso_file"`
+	// Bus type and bus index that the ISO will be mounted on. Can be `ideX`,
+	// `sataX` or `scsiX`.
+	// For `ide` the bus index ranges from 0 to 3, for `sata` from 0 to 5 and for
+	// `scsi` from 0 to 30.
+	// Defaults to `ide2`
+	ISODevice string `mapstructure:"iso_device"`
```
lbajolet-hashicorp, line 35: I wonder if this should just be an `ISODeviceType`, so they are all appended to the list of the appropriate type without needing to give it an index?

builder/proxmox/iso/config.go
```diff
@@ -68,6 +78,90 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, []string, error) {
...
+			case "ide":
+				ideCount++
+			case "sata":
+				sataCount++
+			case "scsi":
+				scsiCount++
+			}
+		}
+		if !(device == "ide" || device == "sata" || device == "scsi") {
+			errs = packersdk.MultiErrorAppend(errs, errors.New("iso_device must be of type ide, sata or scsi. VirtIO not supported for ISO devices"))
```
lbajolet-hashicorp, line 134: This could be the `default` for the `switch device` above I think?

builder/proxmox/iso/config.go
```diff
@@ -68,6 +78,90 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, []string, error) {
...
+		switch device {
+		case "ide":
+			ideCount++
+		case "sata":
+			sataCount++
+		case "scsi":
+			scsiCount++
+		}
+		for idx, iso := range c.AdditionalISOFiles {
+			if iso.Device == c.ISODevice {
```
lbajolet-hashicorp, line 117: Does that `iso.Device` come from proxmox, or is it coming from the SDK? Wondering if this should also be updated to only specify the device type so the plugin can handle the numbering automatically
Of course if this comes from the SDK and we already support it this way, we cannot remove that capability on a minor/patch version, please let me know if this is the case, and we can ignore/revisit my suggestion then.

builder/proxmox/common/config.go
lbajolet-hashicorp: Note: this is moved to `iso/config.go`, might be good enough, assuming the `clone` builder is not also using this? Otherwise we risk this not picking up on invalid configs for that other builder.

builder/proxmox/iso/config.go
```diff
@@ -68,6 +78,90 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, []string, error) {
...
+	virtIOCount := 0
+
+	if c.ISODevice == "" {
+		// set default ISO boot device ide2 if no value specified
+		log.Printf("iso_device not set, using default 'ide2'")
+		c.ISODevice = "ide2"
+		ideCount++
+		// Pass c.ISODevice value over to common config to ensure device index not used by disks
+		c.ISOBuilderCDROMDevice = "ide2"
+	} else {
```
lbajolet-hashicorp, line 95: Should that be an `else`? If the above `if c.ISODevice == ""` branch executes, we can set the default values and use the same validation logic all the same I believe, this may be safe to let it be the normal execution path.

builder/proxmox/common/step_start_vm.go
```diff
@@ -318,36 +359,51 @@ func generateProxmoxDisks(disks []diskConfig) *proxmox.QemuStorages {
...
-			reflect.
-				// We need to get the pointer to the structure so we can
-				// assign a value to the disk
-				ValueOf(&ideDisks).Elem().
-				// Get the field from its name, each disk's field has a
-				// similar format 'Disk_%d'
-				FieldByName(fmt.Sprintf("Disk_%d", ideCount)).
-				Set(reflect.ValueOf(&dev))
-			ideCount++
+			for {
```
lbajolet-hashicorp, line 362: Bit worried about this infinite loop, is there a chance we can never leave here? At first glance this looks risky but maybe this is fine.

lbajolet-hashicorp (commented):
Hey @mpywell,

Thanks for opening this one!

There's a lot happening in this PR, I've reviewed each commit independently to make it clearer to me what's happening, I left a bunch of comments/questions on the code, please feel free to address them and I'll be back for another round of reviews.

builder/proxmox/iso/config.go
mpywell: iso.Device here is the Device field for an additional_iso_files block (from common's Config struct). This was added to ensure the ISO device in the iso builder's Config struct doesn't conflict with a device assignment in the common builder's Config struct, an example where this could happen from a user's perspective would be having `iso_device = sata0` and `additional_iso_files{ device = sata0 }` configured at the same time.

builder/proxmox/iso/config.go
```diff
@@ -32,8 +32,11 @@ type Config struct {
 	ISODownloadPVE bool `mapstructure:"iso_download_pve"`
 	// If true, remove the mounted ISO from the template
 	// after finishing. Defaults to `false`.
-	UnmountISO      bool `mapstructure:"unmount_iso"`
-	shouldUploadISO bool
+	UnmountISO bool `mapstructure:"unmount_iso"`
+	// Keep CDRom device attached to template if unmounting ISO. Defaults to `false`.
+	// Has no effect if unmount is `false`
+	UnmountKeepDevice bool `mapstructure:"unmount_keep_device"`
```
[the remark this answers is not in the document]
mpywell: The additional_iso_files config is in common, and the proxmox-iso builder's iso device is in iso, so the keep option has been added in both places.

builder/proxmox/common/config.go
```diff
@@ -241,7 +241,10 @@ type additionalISOsConfig struct {
 	// Defaults to `false`
 	ISODownloadPVE bool `mapstructure:"iso_download_pve"`
 	// If true, remove the mounted ISO from the template after finishing. Defaults to `false`.
-	Unmount              bool   `mapstructure:"unmount"`
+	Unmount bool `mapstructure:"unmount"`
+	// Keep CDRom device attached to template if unmounting ISO. Defaults to `false`.
+	// Has no effect if unmount is `false`
+	UnmountKeepDevice    bool   `mapstructure:"unmount_keep_device"`
```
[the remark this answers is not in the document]
mpywell: I was going for cohesion, but I do think KeepCDRomDevice is better so I've rerolled with this naming.

builder/proxmox/common/config.go
```diff
@@ -664,6 +667,9 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 				}
 			}
 		}
+		if disk.AsyncIO != "" && !(disk.AsyncIO == "native" || disk.AsyncIO == "threads" || disk.AsyncIO == "io_uring") {
```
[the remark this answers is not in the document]
mpywell: Looks good! I've added this to the reroll.

builder/proxmox/common/step_start_vm.go
mpywell: Good point, I was relying on the iso/additional_iso_files/disk validation from the builder's Prepare function to avoid overallocation. 
`QemuIdeDisks{} QemuSataDisks{} QemuScsiDisks{} QemuVirtIODisks{}` from the SDK have a fixed number of fields as per the limitations of each bus type in Proxmox, I imagine the loop would panic when it hits a Disk_ field that doesn't exist.
I've added a check at the start of the loop to break when all struct fields are exhausted.

builder/proxmox/iso/config.go
mpywell: I've updated this to be the `default` for `switch device` in the reroll commit

builder/proxmox/common/step_start_vm.go
lbajolet-hashicorp: More than a panic I was worried the plugin could be stuck in this loop, which is not trivial to debug when it happens in the wild.
Panics are fine imho, especially if the panic occurs because of a logic bug from the `Prepare`, we can investigate and fix them then. Silent continuations like those introduced with that reroll are more insidious in my experience, as users barely know why the plugin tells them `Max 4 devices` in the verbose logs and doesn't error, but they'll wonder why their device doesn't show up in their configs.
I would maybe suggest returning an error here if we bust that limit, panicking might be suboptimal because of the stacktrace that is returned (although for debugging this is kinda good, and it explicitly tells users to report the problem), and silently continuing feels wrong to me.

builder/proxmox/iso/config.go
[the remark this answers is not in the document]
lbajolet-hashicorp: OK I didn't understand that the option was to point at different devices.
So just to be clear, is the `additionalISOConfig` squashed into the config, or is it imported as a block for the ISO builder? If it is squashed, we'll have a conflict between the two options.

EDIT: looked at the code, looks like it'll be fine since the `additionalISOConfig` objects are added as a slice to the configuration, so each object will not be squashed into the config directly, both options won't conflict! We can keep them both, looks legit to me

builder/proxmox/iso/config.go
lbajolet-hashicorp: Yeah regarding disks I can see a reason why we'd want to keep the order in which we enumerate them (even when modern OSes don't access their devices by the order they're discovered but through UUID), regarding cdrom devices, I believe they can fill-in the remaining slots, that would be likely be simpler and probably won't break existing configs.
What do you think?

builder/proxmox/iso/config.go
lbajolet-hashicorp: Gotcha, so this is indeed part of the proxmox plugin, so we can change this for this plugin. I would maybe revisit the way we add those iso devices to the VM config, using a single function from `common`, this way we can use it from both the `Clone` and `ISO` builders, and not have this logic scattered around.
Maybe we can add this as a step is relevant? Otherwise a function should do the trick possibly.
Let me know what you think about this idea, and if you need some help with the code, feel free to ask and I can take a jab at that!

builder/proxmox/iso/config.go
mpywell: I tracked down the original additional_iso_files issue https://github.com/hashicorp/packer/issues/7950 and PR https://github.com/hashicorp/packer/pull/9653 that brought the logic into the builder to understand why a device and index are specified. It looks like the builder at the time only accepted one set of iso device config (what is now the proxmox-iso builder) and additional_iso_files was added in.

It seems the device and index weren't explicit requirements in the feature request, so we could look at removing the index attribute for all ISOs and just enable the assignment of a device type. Looking at https://github.com/hashicorp/packer/pull/9653#discussion_r467550534 we've ended up implementing the second proposed option for disks to meet the new API struct format anyway so we would just be folding in ISO devices.

It would certainly simplify aspects of this PR 😉 however this would be a breaking change for existing packer configs where an additional_iso_files block has a device field value.

I'm happy to have a go at updating everything to use device type assignments without indexes.. with the above information do you think this is a viable way forward?

builder/proxmox/iso/config.go
mpywell: With the background obtained answering the above comment on device indexes, I think iso and additional_iso_files could potentially be merged together and maybe the packer config adapted to just have 1 or more `iso` blocks rather than `iso` fields/values outside of a block for the first ISO then multiple `additional_iso_files` blocks for each after? They could be processed in the order they're defined as the disks blocks are, that way you could consistently define the first ISO as an Operating System installer media, second is an autounattend.xml for Windows builds (just examples here), third VirtIO drivers and so on. 
I do then wonder what the point of the iso builder would be if all the ISO logic moves to common.

builder/proxmox/common/step_start_vm.go
mpywell: Yeah I made the wrong call here 😅 I can update this to be a ui response rather than a log.Print

mpywell:
Hey @lbajolet-hashicorp,

I've pushed a reroll that consolidates iso builders iso and additional_iso_files into a single common iso struct.

```
  # HCL v1.8 iso builder example
  # Define boot iso device
  iso_url          = "https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/debian-12.5.0-amd64-netinst.iso"
  iso_checksum     = "sha512:33c08e56c83d13007e4a5511b9bf2c4926c4aa12fd5dd56d493c0653aecbab380988c5bf1671dbaea75c582827797d98c4a611f7fb2b131fbde2c677d5258ec9"
  iso_storage_pool = var.iso_storage_pool
  unmount_iso      = true

  # define additional iso device
  additional_iso_files {
    device           = "ide0"
    iso_storage_pool = var.iso_storage_pool
    unmount          = true
    iso_url          = "https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/stable-virtio/virtio-win.iso"
    iso_checksum     = "none"
  }

  # HCL for this commit
  isos {
    type         = "ide"
    storage_pool = var.iso_storage_pool
    unmount      = true
    iso_url      = "https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/debian-12.5.0-amd64-netinst.iso"
    iso_checksum = "sha512:33c08e56c83d13007e4a5511b9bf2c4926c4aa12fd5dd56d493c0653aecbab380988c5bf1671dbaea75c582827797d98c4a611f7fb2b131fbde2c677d5258ec9"
  }
  isos {
    type         = "ide"
    storage_pool = var.iso_storage_pool
    unmount      = true
    iso_url      = "https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/stable-virtio/virtio-win.iso"
    iso_checksum = "none"
  }
```
The isos are enumerated in the order they are defined just as disks are, after disks have been enumerated.

This enables all ISOs to have uniform configuration and makes common the central point of storage validation for both the iso and clone builders.

I've also added a prestep to the clone builder to map source vm disks to their device type and index before common enumerates packer defined disks and isos to enable the appending of new disks and isos to a cloned vm.

generateProxmoxDisks now returns ui and state errors if device types are overallocated

lbajolet-hashicorp:
Hey @mpywell,

Just to let you know, I haven't forgotten about this PR, I plan on reviewing it this week (FYI I'll be OOO for a couple weeks after this so this might take a while before getting folded into a release, sorry about this).

I like the new look for the template, reifying everything as a list of blocks seems on point to me.

I'll take a look at the code to make sure this makes sense.

Question: this changes the supported grammar, are we expecting a breaking change in the configs?

mpywell:
Hey @lbajolet-hashicorp,

Yeah at a functional level there are no differences between the main ISO and the additional ISOs, ie. both are processed by the proxmox-api-go SDK in the same way and accept the same configuration options in the Proxmox backend. The Proxmox backend also only has the one 'cdrom' VM device for attaching ISOs.

I pivoted to making the ISO definition the same for both based on this and because of the earlier challenges around validating two types of ISO config, but it does also mean that packer configs for the next release would need adjusting to this now iso block format, the 1.8 values wouldn't be accepted anymore.

I didn't quite understand your comment about making the main ISO config an attribute that accepts the block, could you show me in a code snippet example? Or feel free to add a commit 😄

lbajolet-hashicorp:
Hey @mpywell,

Thanks for the quick response, much appreciated!

Regarding the main ISO config being an attribute instead of an entry to the ISO block list I meant something like this:

```hcl
  ## HCL for current HEAD
  isos {
    type         = "ide"
    storage_pool = var.iso_storage_pool
    unmount      = true
    iso_url      = "https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/debian-12.5.0-amd64-netinst.iso"
    iso_checksum = "sha512:33c08e56c83d13007e4a5511b9bf2c4926c4aa12fd5dd56d493c0653aecbab380988c5bf1671dbaea75c582827797d98c4a611f7fb2b131fbde2c677d5258ec9"
  }
  isos {
    type         = "ide"
    storage_pool = var.iso_storage_pool
    unmount      = true
    iso_url      = "https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/stable-virtio/virtio-win.iso"
    iso_checksum = "none"
  }

  ## HCL example with ISO being an attribute
  iso = {
    type         = "ide"
    storage_pool = var.iso_storage_pool
    unmount      = true
    iso_url      = "https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/debian-12.5.0-amd64-netinst.iso"
    iso_checksum = "sha512:33c08e56c83d13007e4a5511b9bf2c4926c4aa12fd5dd56d493c0653aecbab380988c5bf1671dbaea75c582827797d98c4a611f7fb2b131fbde2c677d5258ec9"
  }
  additional_isos = [{
    type         = "ide"
    storage_pool = var.iso_storage_pool
    unmount      = true
    iso_url      = "https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/stable-virtio/virtio-win.iso"
    iso_checksum = "none"
  }]
```

This can be done by playing with mapstructure tags in the config, typically for the ISO builder this could be something like this:

```go
type Config struct {
	[...]
	ISO ISOsConfig `mapstructure:"iso" required:"true"`
	// ISO files attached to the virtual machine.
	// See [ISO Files](#iso-files).
	ISOs []ISOsConfig `mapstructure:"additional_iso_files"`
}
```

Note: I renamed your `isos` to `additional_iso_files` here in order to maintain compatibility, I think we should strive to avoid breaking existing configs, unless when calculated.

The ISO specified as `iso` could be prepended to the ISO list, and then passed to the common logic that adds ISOs to the VMConfig on proxmox.

This has the benefit of explicitly stating that this ISO is a special case that is only available for the ISO builder (clone only supports `additional_iso_files` anyway I think?), and is clearly marked as such.
To be fair, your approach works great as well, I think it maybe should be documented that the first ISO is meant to be the boot one? Not sure if the order changes something (my guess is that'd depend on the EFI/BIOS config/revision, and the OS after that), but if we keep the current approach we should still document it at least I think.

I would think we can also preserve backwards compatibility (albeit maybe with less configuration possibilities) by keeping the original main ISO config schema alive (with deprecation warnings), and build an `ISOsConfig` from those, so we can still keep the updated logic in one spot.
If we go with this, we should also make sure the two alternatives are mutually exclusive, so that means updating `Prepare` to do that check.

What do you think? Let me know if something's unclear, happy to elaborate more if needed.
And please feel free also to tell me if you want me to delve into the code, I've got bandwidth for that a bit, so happy to tinker if you think that'd help.

Thanks again!

mpywell:
Hey @lbajolet-hashicorp,

I've pushed another reroll with your suggestion, the boot iso is prepended to common's ISO slice and enumerated through common as before.

 ```
  iso = {
    type         = "ide"
    iso_storage_pool = var.iso_storage_pool
    unmount      = true
    iso_url      = "https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/debian-12.5.0-amd64-netinst.iso"
    iso_checksum = "sha512:33c08e56c83d13007e4a5511b9bf2c4926c4aa12fd5dd56d493c0653aecbab380988c5bf1671dbaea75c582827797d98c4a611f7fb2b131fbde2c677d5258ec9"
  }
  additional_iso_files = [{
    type         = "ide"
    iso_storage_pool = var.iso_storage_pool
    unmount      = true
    iso_url      = "https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/stable-virtio/virtio-win.iso"
    iso_checksum = "none"
  }]
```

Note: As we're moving to enumeration for ISO devices the additional_iso_files block's type field doesn't accept device indexes anymore, so there will still be changes required in packer configs. I might try to clarify this in docs through another commit

**Comment by mpywell** [comment]
Hey @lbajolet-hashicorp just following up to see when we might be able to pick this one up again?

**Review by lbajolet-hashicorp — COMMENTED** [review]
Hi @mpywell,

Thanks for your persistence, and apologies for not reviewing this earlier.

Overall the code looks good to me, I am however worried about this change being breaking for existing configurations. We could make this part of a major release if there's no way around it, as the code does fix several issues, and IMHO makes the management of extra disks/ISOs much clearer, but I would like to explore if this needs to be a breaking change.

My understanding is that `ISOsConfig` replaces the current `additionalISOsConfig`, but I wonder if we can make the transition easier for the future.
Typically is there a way to convert entries of the current structure into the new ones. My rationale being that if an existing template defines `additional_iso_files` in their configuration, we can maybe convert those into the new structure, even if it doesn't map 1:1; typically device contains both how the media is attached and an ID that is now computed on the fly, we could maybe extract the device type from `additional_iso_files` and use the new numbering strategy as a replacement. This would allow existing templates to conform to the new model, even if there's a slight change. Or we could try to fit the numbering at the same time if this is important for some reason.
This would be accompanied by a warning if using the old structure, so users know they should update their template to match the new model, making the transition between the old world and the new one smoother.

I'll rely on you to tell me if this is possible or not, and if you want to meet to discuss that, please let me know and we can probably schedule something; I suspect this'll be easier for us to reach an understanding, but we can also continue discussing this on the PR.

Thanks again for the effort, we really appreciate this.

**Comment by mpywell** [comment]
Hi @lbajolet-hashicorp, thanks for your feedback.

I think it should be possible to implement a conversion for existing `additionalISOsConfig` configs to `ISOsConfig`, leave it with me and I'll see what I can do, it shouldn't take too long.

If I have any difficulty I'll reach out and take you up on organising a meet.

**Comment by mpywell** [comment]
Hi @lbajolet-hashicorp,

I've added backwards compatibility support for all of the ISO changes, let me know your thoughts.

File: builder/proxmox/common/config.go [hunk 11]
```diff
@@ -636,12 +666,113 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 		log.Printf("OS not set, using default 'other'")
 		c.OS = "other"
 	}
-	ideCount := 0
-	sataCount := 0
-	scsiCount := 0
-	virtIOCount := 0
+	// validate iso devices
+	for idx := range c.ISOs {
```
lbajolet-hashicorp, line 670: Doesn't need to happen in this PR, but I wonder if we can make each `ISOs` here have its own `Prepare` function with a similar signature as the global `Prepare`, if only to split the logic into something more bite-sized.

builder/proxmox/common/config_test.go
```diff
@@ -137,24 +138,24 @@ func TestVMandTemplateName(t *testing.T) {
...
+		expectedToFail bool
+		ISOs           map[string]interface{}
 	}{
 		{
 			name:           "missing ISO definition should error",
 			expectedToFail: true,
-			additionalISOFiles: map[string]interface{}{
-				"device": "ide1",
+			ISOs: map[string]interface{}{
+				"type": "ide",
```
lbajolet-hashicorp, line 151: I wonder if we should keep (at least some of) those tests as-is just to validate the current logic doesn't break with the last changes, we can add some with type alone (or type+index)

lbajolet-hashicorp: Note: I see we're introducing deprecated option conversion tests below, might be worth moving over some of those tests over there?

builder/proxmox/common/config_test.go
```diff
@@ -244,7 +245,47 @@ func TestAdditionalISOs(t *testing.T) {
...
+func TestDeprecatedISOOptionsAreConverted(t *testing.T) {
+	isotests := []struct {
+		name string
+		ISOs map[string]interface{}
+	}{
+		{
+			// Ensure deprecated device field is converted
+			name: "device should be converted to type and index",
+			ISOs: map[string]interface{}{
+				"device":   "ide3",
```
lbajolet-hashicorp, line 259: Might be superfluous, but I wonder if we should add cases with a device index too high? These are probably caught by the blob above though, I'll leave it up to you to decide

lbajolet-hashicorp (commented):
Hey @mpywell,

Thanks for the reroll!
From the looks of it, I would think we are good in terms of backwards-compatibility, and the new features, overall LGTM!

I left a few last nits/questions on the code, feel free to address those when you can, but we're nearing a merge, kudos to you!

builder/proxmox/common/config_test.go
mpywell: Good point, I've copied some of the TestISOs cases over to TestDeprecatedISOOptionsAreConverted using the `device` field

builder/proxmox/common/config_test.go
mpywell: The handling of bus overallocation moved to `generateProxmoxDisks` in `builder/proxmox/common/step_start_vm.go` because there's also the potential for overallocation when preserving clone builder disks.

I've pushed an update for `TestGenerateProxmoxDisks` in `builder/proxmox/common/step_start_vm_test.go` to test overallocating sata storage with a mix of defined disks, ISOs and clone source vms mappings, do you think one is needed for each storage type?

mpywell (commented):
Hey @lbajolet-hashicorp, I've pushed a reroll that I think addresses some of the points raised, let me know if you have any further questions

lbajolet-hashicorp (approved):
Left a final couple of nits, but overall LGTM!

Thanks for pushing that one @mpywell, much appreciated.
I'll let you address those as you see fit, but imho we should be ready to merge after that.

I'm pre-approving this PR so it's not a blocker later

builder/proxmox/common/config_test.go
```diff
@@ -284,6 +317,15 @@ func TestDeprecatedISOOptionsAreConverted(t *testing.T) {
 			if config.ISOs[0].Index != index {
 				t.Errorf("Expected device to be converted to index %s", index)
 			}
+
+			if c.expectedToFail == true && err == nil {
```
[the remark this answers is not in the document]
mpywell: Fair, I copied those conditions from `TestISOs` 239-245, updated there as well.

mpywell:
@lbajolet-hashicorp I've resolved the last few items, should be right to merge

lbajolet-hashicorp:
Awesome, thanks for the update and the ping @mpywell, much appreciated.

We are good to go, merging this one now, and I'll schedule a release for the plugin ASAP, we've got a few items to cram into this one :)

PR #269, opened 2024-06-18
[AUTO] Update question.md template to remove reference to Google Groups mailing list.

lbajolet-hashicorp:
Since the mailing list is not usable anymore on Google groups, we
remove the link to it in our question template.


nywilken (approved):
LGTM

PR #274, opened 2024-07-30
fix for remove the old cloud-init on local disks

### LINKED ISSUE #273: Packer did not remove the old cloud-init on local disks
When creating an os+db image, Packer throws an error: Error updating template: ide1 - cloud-init drive is already attached at 'ide0'.

The reason is that Packer did not remove the old cloud-init and trying to insert a new one.

The removal occurs in the code: https://github.com/hashicorp/packer-plugin-proxmox/blob/main/builder/proxmox/common/step_remove_cloud_init_drive.go#L56

If you create cloud-init on local disks, the file name will also include the disk format:

     my $scfg = PVE::Storage::storage_config($storecfg, $storeid);
            my $name = "vm-$vmid-cloudinit";

            my $fmt = undef;
            if ($scfg->{path}) {
                $fmt = $disk->{format} // "qcow2";
                $name .= ".$fmt";
            } else {
                $fmt = $disk->{format} // "raw";
            }

VadimLakt:
When creating an os+db image, Packer throws an error: Error updating template: ide1 - cloud-init drive is already attached at 'ide0'.

The reason is that Packer did not remove the old cloud-init and trying to insert a new one.

If you create cloud-init on local disks, the file name will also include the disk format:

 my $scfg = PVE::Storage::storage_config($storecfg, $storeid);
        my $name = "vm-$vmid-cloudinit";

        my $fmt = undef;
        if ($scfg->{path}) {
            $fmt = $disk->{format} // "qcow2";
            $name .= ".$fmt";
        } else {
            $fmt = $disk->{format} // "raw";
        }

Closes #273

### DIFF
--- a/builder/proxmox/common/step_remove_cloud_init_drive.go
+++ b/builder/proxmox/common/step_remove_cloud_init_drive.go
@@ -53,7 +53,7 @@ func (s *stepRemoveCloudInitDrive) Run(ctx context.Context, state multistep.Stat
 	}
 
 	for _, controller := range diskControllers {
-		if vmParams[controller] != nil && strings.Contains(vmParams[controller].(string), "-cloudinit,media=cdrom") {
+		if vmParams[controller] != nil && strings.Contains(vmParams[controller].(string), "-cloudinit") && strings.Contains(vmParams[controller].(string), ",media=cdrom") {
 			delete = append(delete, controller)
 		}
 	}


mizhka:
Hi, 

Any progress on it?

lbajolet-hashicorp (commented):
Hi @VadimLakt,

Thanks for the PR! The code looks good to me, but before merging may I ask how this change fixes the issue at hand?
It's not clear to me what the controller string looks like in the case you describe, I assume there's extra data in there between the cloudinit and the media and that's why it fails with the current check, but I would appreciate if you cloud share what kind of string makes the current check fail and how the new code fixes that.

Thanks again!

VadimLakt:
Hi @lbajolet-hashicorp,

Thank you for your response. 

I apologize for my delayed reply; we had difficulty accessing the problematic server. I added some debugging to the issue area, and the line that fails the check looks like this: var: vmParams[controller] value: "RAID5:154/vm-154-cloudinit.qcow2,media=cdrom,size=4M". 

Between "cloudinit" and "media," there is also the disk format "qcow2." 

If you need any further adjustments, feel free to let me know!

lbajolet-hashicorp (approved):
Hi again @VadimLakt,

Thanks for the update, overall this looks good to me and if it unblocks your use-case I don't see a reason why we should reject it. The old strings are still detected as they should from what I can see.
I would think adding a couple of tests to make sure it does would be best, but given the triviality of the change, I think we can do without for now.

I'll merge this right now, thanks again for the contribution!

PR #283, opened 2024-09-28
[Issue 71] add skip_convert_to_template option

### LINKED ISSUE #71: Add "convert_to_template"  option to iso and clone builders

mpywell:
Rewrite of #72 on current codebase.

Implements config option skip_convert_to_template which skips converting the build to a Template and leaves the build artifact as a VM.

Closes #71

### DIFF
--- a/builder/proxmox/common/step_finalize_config.go
+++ b/builder/proxmox/common/step_finalize_config.go
@@ -23,6 +23,7 @@ type finalizer interface {
 	GetVmConfig(*proxmox.VmRef) (map[string]interface{}, error)
 	SetVmConfig(*proxmox.VmRef, map[string]interface{}) (interface{}, error)
 	StartVm(*proxmox.VmRef) (string, error)
+	ShutdownVm(*proxmox.VmRef) (string, error)
 }
 
 var _ finalizer = &proxmox.Client{}
@@ -46,7 +47,7 @@ func (s *stepFinalizeConfig) Run(ctx context.Context, state multistep.StateBag) 
 
 	vmParams, err := client.GetVmConfig(vmRef)
 	if err != nil {
-		err := fmt.Errorf("error fetching template config: %s", err)
+		err := fmt.Errorf("error fetching config: %s", err)
 		state.Put("error", err)
 		ui.Error(err.Error())
 		return multistep.ActionHalt
@@ -140,6 +141,17 @@ func (s *stepFinalizeConfig) Run(ctx context.Context, state multistep.StateBag) 
 	changes["delete"] = strings.Join(deleteItems, ",")
 
 	if len(changes) > 0 {
+		// Adding a Cloud-Init drive or removing CD-ROM devices won't take effect without a power off and on of the QEMU VM
+		if c.SkipConvertToTemplate {
+			ui.Say("Hardware changes pending for VM, stopping VM")
+			_, err := client.ShutdownVm(vmRef)
+			if err != nil {
+				err := fmt.Errorf("Error converting VM to template, could not stop: %s", err)
+				state.Put("error", err)
+				ui.Error(err.Error())
+				return multistep.ActionHalt
+			}
+		}
 		_, err := client.SetVmConfig(vmRef, changes)
 		if err != nil {
 			err := fmt.Errorf("Error updating template: %s", err)
@@ -151,7 +163,7 @@ func (s *stepFinalizeConfig) Run(ctx context.Context, state multistep.StateBag) 
 
 	// When build artifact is to be a VM, return a running VM
 	if c.SkipConvertToTemplate {
-		ui.Say("skip_convert_to_template set, resuming VM")
+		ui.Say("Resuming VM")
 		_, err := client.StartVm(vmRef)
 		if err != nil {
 			err := fmt.Errorf("Error starting VM: %s", err)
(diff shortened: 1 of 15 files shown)


builder/proxmox/common/step_finalize_config.go
```diff
@@ -17,18 +17,19 @@ import (
 // stepFinalizeTemplateConfig does any required modifications to the configuration _after_
```
lbajolet-hashicorp, line 17: Minor nit, I'd suggest updating the documentation here just so the name of the step is correctly reflected.

builder/proxmox/common/step_finalize_config.go
```diff
@@ -17,18 +17,19 @@ import (
...
 // the VM has been converted into a template, such as updating name and description, or
 // unmounting the installation ISO.
-type stepFinalizeTemplateConfig struct{}
+type stepFinalizeConfig struct{}
 
-type templateFinalizer interface {
+type finalizer interface {
 	GetVmConfig(*proxmox.VmRef) (map[string]interface{}, error)
 	SetVmConfig(*proxmox.VmRef, map[string]interface{}) (interface{}, error)
+	StartVm(*proxmox.VmRef) (string, error)
```
lbajolet-hashicorp, line 25: Is there a reason to add `StartVm` to that contract? It seems that the function is not invoked in the step, and AFAICT this is not used elsewhere, so I wonder if that might be a leftover from a WIP state of the PR?
Feel free to correct me if I'm wrong though, I might've missed something

lbajolet-hashicorp (dismissed):
Overall LGTM, thanks @mpywell

Just left a very small nit and a question about a mock signature changing, but overall the code looks good to me, once you've had a chance to address my comments we should be good to go!

Also apologies for leaving this one aside for so long, I'll do a full review of the open PRs on Proxmox today, with the aim of scheduling a release for later this week if possible.

Thanks again!

builder/proxmox/common/step_finalize_config.go
mpywell: It was left over from a WIP version of the PR where the returned artifact was a running VM. I noted in https://github.com/hashicorp/packer-plugin-proxmox/pull/72#issuecomment-2022956177 a desire to have a running VM returned, so I've rerolled with those changes.

mpywell:
Hi @lbajolet-hashicorp, I've rerolled and addressed comments

builder/proxmox/common/step_finalize_config.go
lbajolet-hashicorp: Ah yeah understood, I missed the resume VM part.

I wonder though, the current workflow appears to work (can't test it as I don't have a proxmox machine on hand), but I would think we can keep the VM running throughout the process if we're not exporting to a template, don't we?
In this case we can change a bit the `Cleanup` method of the `step_start_vm` so it doesn't kill and delete the VM, unless it cannot?
Also to be sure, since the cleanup for the VM will run after `step_finalize_config`, will the VM still exist? I'd assume this will be stopped and removed, won't it?

mpywell: A stop and start of the VM is required if there are VM hardware changes for the VM's Cloud-Init storage and CD-ROM devices (step `step_finalize_config`).

These changes aren't picked up by the VM unless it's completely stopped then started again, they can't be hot-swapped or mounted on an ACPI reboot. This is a limitation on the Proxmox backend.

I've reworked the logic to only stop the VM artifact if there are changes to be done in `step_finalize_config` (if len(changes) > 0) that way if Cloud-Init isn't being configured or CD-ROM devices are being kept, the VM won't restart.

mpywell:
Hey @lbajolet-hashicorp just checking in to see if anything further is needed for this one?

builder/proxmox/common/step_finalize_config.go
```diff
@@ -139,6 +141,17 @@ func (s *stepFinalizeTemplateConfig) Run(ctx context.Context, state multistep.St
 	changes["delete"] = strings.Join(deleteItems, ",")
 
 	if len(changes) > 0 {
+		// Adding a Cloud-Init drive or removing CD-ROM devices won't take effect without a power off and on of the QEMU VM
+		if c.SkipConvertToTemplate {
```
lbajolet-hashicorp, line 145: Out of curiosity, why is the shutdown only done when there are changes? I would've thought this should be coupled with the call to `StartVM` in every case should you decide not to convert the VM to a template.
One scenario that I have in mind (no clue if that is realistic or not though), if we don't have changes the VM won't be shutdown, and we'll try to start it again afterwards, could the call fail if the VM is still running?

lbajolet-hashicorp (commented):
Hey @mpywell,

Apologies I took so long to revisit this PR; I left one last comment on the stop/start VM code, I understand that you don't want to restart the VM unnecessarily, but it seems to me that trying to start a VM if it wasn't stopped might fail, but please feel free to correct me if that's not the case.

Besides this nit/comment, the PR looks good to me, we should be ready to merge soon, thanks!

Gauvino:
Can this be merged now or not ?

b-bittner:
This would be a great option to be added. Really looking forward for it!

tanmay-hc:
Closing this for https://github.com/hashicorp/packer-plugin-proxmox/pull/356

PR #287, opened 2024-10-04
[Issue 276] QEMU Agent configuration settings

### LINKED ISSUE #276: Agent type selection (virtIO or ISA)
#### Community Note

#### Description
- Packer can't communicate with an OpenBSD VM because the agent type isn't set to ISA:

image

- Current available option (agent) handle only boolean.

#### Use Case(s)
Add an option to choose the agent type. Or upgrade the agent option to handle the type.

- OpenBSD system works with ISA
- VM creation by Packer have to permit to select the agent type (builder.pkr.hcl)
- This way Packer can send signals and retrieve informations from an OpenBSD system as it does for other systems working with virtIO

#### Potential configuration

```
```

#### Potential References
[Proxmox ISO builder Options](https://developer.hashicorp.com/packer/integrations/hashicorp/proxmox/latest/components/builder/iso#optional:):
`qemu_agent (boolean) - Enables QEMU Agent option for this VM. When enabled, then qemu-guest-agent must be installed on the guest.`

[Proxmox CLI Options](https://pve.proxmox.com/pve-docs/qm.1.html):
` --agent [enabled=]<1|0> [,freeze-fs-on-backup=<1|0>] [,fstrim_cloned_disks=<1|0>] [,type=<virtio|isa>] `

[Proxmox API Options](https://pve.proxmox.com/pve-docs/api-viewer/#/nodes/{node}/qemu/{vmid}/config):
image

mpywell:
Extend QEMU agent handling by introducing the `qemu_quest_agent` block

`qemu_guest_agent` adds the ability to set each of the QEMU Guest Agent's available settings:
- Enable/Disable the agent (replaces the `qemu_agent` bool)
- Setting the agent type
- Toggling freeze/thaw filesystems on backup
- Toggling fstrim after disk move or VM migration

Each added setting defaults to Proxmox's backend defaults when not configured.

Closes #276

### DIFF
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -243,8 +243,8 @@ type agentConfig struct {
 	Enabled config.Trilean `mapstructure:"enabled"`
 	// Sets the Agent Type. Must be `isa` or `virtio`. Defaults to `virtio`
 	Type string `mapstructure:"type"`
-	// Enable freeze/thaw of guest filesystem on backup. Defaults to `true`
-	Freeze config.Trilean `mapstructure:"freeze"`
+	// Disable freeze/thaw of guest filesystem on backup. Defaults to `false`
+	DisableFreeze bool `mapstructure:"disable_freeze"`
 	// Run guest-trim after a disk move or VM migration. Defaults to `false`
 	FsTrim bool `mapstructure:"fstrim"`
 }
@@ -646,20 +646,15 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 		warnings = append(warnings, "proxmox is deprecated, please use proxmox-iso instead")
 	}
 
-	// Default qemu_agent to true
-	if c.Agent != config.TriFalse {
+	if c.Agent != config.TriUnset {
 		warnings = append(warnings, "qemu_agent is deprecated and will be removed in a future release. define QEMU agent settings in a qemu_guest_agent block instead")
 		// convert to qemu_guest_agent block value
-		c.GuestAgent.Enabled = config.TriTrue
+		c.GuestAgent.Enabled = c.Agent
 	}
 
 	// Default qemu_guest_agent.enable to true
 	if c.GuestAgent.Enabled != config.TriFalse {
 		c.GuestAgent.Enabled = config.TriTrue
-	}
-	// Default qemu_guest_agent.freeze to true
-	if c.GuestAgent.Freeze != config.TriFalse {
-		c.GuestAgent.Freeze = config.TriTrue
 	}
 
 	switch c.GuestAgent.Type {
--- a/builder/proxmox/common/config.go
+++ b/builder/proxmox/common/config.go
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: MPL-2.0
 
 //go:generate packer-sdc struct-markdown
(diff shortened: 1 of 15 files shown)


builder/proxmox/common/config.go
```diff
@@ -606,7 +648,27 @@ func (c *Config) Prepare(upper interface{}, raws ...interface{}) ([]string, []st
 
 	// Default qemu_agent to true
 	if c.Agent != config.TriFalse {
-		c.Agent = config.TriTrue
+		warnings = append(warnings, "qemu_agent is deprecated and will be removed in a future release. define QEMU agent settings in a qemu_guest_agent block instead")
```
lbajolet-hashicorp, line 651: I would suggest we only print that warning if `qemu_agent` is unset here, not if it's not explicitly `false`, otherwise we'll see it all the time unless we add `qemu_agent = false` to the template.

builder/proxmox/iso/config_test.go
```diff
@@ -103,8 +103,8 @@ func TestBasicExampleFromDocsIsValid(t *testing.T) {
 	if b.config.Disks[0].CacheMode != "none" {
 		t.Errorf("Expected disk cache mode to be 'none', got %s", b.config.Disks[0].CacheMode)
 	}
-	if b.config.Agent.True() != true {
-		t.Errorf("Expected Agent to be true, got %t", b.config.Agent.True())
+	if b.config.GuestAgent.Enabled.True() != true {
```
lbajolet-hashicorp, line 106: I would suggest simplifying this one; 
```suggestion
	if !b.config.GuestAgent.Enabled.True() {
```

lbajolet-hashicorp (dismissed):
Thanks for this one @mpywell!

Overall LGTM, I left a couple of suggestions, feel free to address them and I'll come back for another round of review after that.

builder/proxmox/common/config.go
mpywell: It occurred to me that we don't need to set a default value for `c.Agent` given the replacement `c.GuestAgent.Enabled` has a default set further down, so I've changed the logic to warn and convert the value only if the `c.Agent` is not Unset.

builder/proxmox/common/config.go
```diff
@@ -207,6 +210,45 @@ type Config struct {
...
+// ```
+type agentConfig struct {
+	// Enable QEMU Agent option for this VM. When enabled
+	// `qemu-guest-agent` must be installed on the guest. When disabled
+	// `ssh_host` should be used. Defaults to `true`.
+	Enabled config.Trilean `mapstructure:"enabled"`
+	// Sets the Agent Type. Must be `isa` or `virtio`. Defaults to `virtio`
+	Type string `mapstructure:"type"`
+	// Enable freeze/thaw of guest filesystem on backup. Defaults to `true`
+	Freeze config.Trilean `mapstructure:"freeze"`
```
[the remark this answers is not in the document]
mpywell: Makes sense, have rerolled with `disable_freeze` as a bool

builder/proxmox/iso/config_test.go
mpywell: Done, added to reroll

mpywell:
Hi @lbajolet-hashicorp have rerolled this one and added comments in the review

DerLinkman:
This would be really nice to have merged :)

dreibh:
Indeed, this pull request would be useful. Currently, for building OpenBSD or NetBSD VMs, it requires to manually set the agent type to ISA.

PR #297, opened 2024-10-30
[Issue 277] add cloud_init_disable_upgrade_packages option

### LINKED ISSUE #277: Option for the Cloud-Init "Upgrade Packages" feature
#### Community Note

#### Description

Proxmox VE 8 introduced a cloud-init option called `ciupgrade` that controls wheter a machine should upgrade it's packages on boot. Currently this option cannot be controlled with Packer. 

#### Use Case(s)

As the default value of this option is set to `true`this means that when using the Proxmox Cloud-Init (i.e. with the `ipconfig` Packer option or other) the Cloud-Init drive gets created which then in return, by default, upgrades the packages on the first boot. This can lead to unexpected behaviour when running different workflows after the VM creation (apt in locked state etc.). 

#### Potential configuration

A boolean flag called `upgrade_packages` with the default set to `true`, to match the API default, that can be overriden in the builder template.

#### Potential References
Documentation:
https://pve.proxmox.com/wiki/Cloud-Init_Support

Feature Introduction:
https://pve.proxmox.com/wiki/Roadmap#Proxmox_VE_8.0

mpywell:
Add option to disable the `Upgrade Packages` setting on Cloud-Init Drives attached to VMs.

`Upgrade Packages` is introduced in Proxmox VE 8 and logic is added to only configure this option on backend versions that are compatible.

Closes #277


CaptaiNiveau:
Just started implementing this myself, until I stumbled upon this PR. @JenGoldstrich [you mentioned over here](https://github.com/hashicorp/packer-plugin-proxmox/pull/308#pullrequestreview-3064628944) that you wanted to discuss maintenance status of this plugin with the team - what's the current state of that? It'd be nice if small things like this PR were approved.

I need this option, because otherwise Arch Linux automatically upgrades its packages in the iso and immediately fills up its cowspace. Yes, I can increase the size of the cowspace afterwards, but just disabling auto upgrade is obviously the significantly better fix.

Thanks for all the nice IaC tooling, would love to see it improve :)

CaptaiNiveau:
Oh btw, I rebased that PR and fixed the small merge conflicts [over here](https://github.com/CaptaiNiveau/packer-plugin-proxmox/tree/feat/277_cloud_init_upgrade_packages). I'll be testing it out in the next few days but I doubt that there will be any issues.

CaptaiNiveau:
Hm, sadly it doesn't seem to work as expected yet. Just installed the plugin, but swapping it on/off doesn't change anything about the `upgrade packages` flag in Proxmox Cloud Init. Hm, I'll try to debug this.

CaptaiNiveau:
Ah, that PR actually only focused on the resulting template. I expanded on it and also use this variable if the vm is cloned instead of the iso. That behavior is in line with other existing variables, like the ssh public key. That one is also pulled in from the common communicator module into the Cloud Init configuration of the cloned vm.

mpywell:
Hey @CaptaiNiveau good catch, I must have missed adding this for the clone builder back when I submitted this PR 2 years ago. I do hope this repo gets some attention, I lost momentum as each PR I wrote had a long wait before being reviewed, some like this one are still awaiting an initial review.

builder/proxmox/common/step_finalize_config.go
```diff
@@ -22,6 +22,7 @@ type stepFinalizeTemplateConfig struct{}
 type templateFinalizer interface {
 	GetVmConfig(*proxmox.VmRef) (map[string]interface{}, error)
 	SetVmConfig(*proxmox.VmRef, map[string]interface{}) (interface{}, error)
+	Version() (proxmox.Version, error)
```
tanmay-hc, line 26: This introduces a new `version` dependency on every finalize path, even when `cloud_init` is off. That changes existing behavior and can now fail builds unrelated to this option. Can we move the version lookup into the exact branch where we actually need to decide whether to write `ciupgrade`?

builder/proxmox/common/step_finalize_config.go
```diff
@@ -87,11 +96,25 @@ func (s *stepFinalizeTemplateConfig) Run(ctx context.Context, state multistep.St
...
+							changes["ciupgrade"] = false
+						case false:
+							changes["ciupgrade"] = true
+						}
+					} else {
+						// only write to UI if the cloud_init_disable_upgrade_packages is configured but the backend is an incompatible version
+						if c.CloudInitDisableUpgradePackages {
+							ui.Say("cloud_init_disable_upgrade_packages is set to true, but not supported in Proxmox versions lower than 8. No changes made.")
+						}
+					}
```
tanmay-hc, line 117: This also changes the default behavior: when the new field is omitted, we still write `ciupgrade=true`. That means existing templates/clones with `ciupgrade=false` can get flipped on upgrade. We probably need tri-state semantics here, or to only send `ciupgrade` when the user explicitly set the option.

tanmay-hc (changes requested):
I know the review is a bit too late. Apologies for that.
Please address the comments and do a rebase.
We should be good to merge.

mpywell:
Hi @tanmay-hc, have rebased and made some adjustments.

- CloudInitDisableUpgradePackages is now Trilean
- Version is only called if the Trilean has a set value
- I've added the functionality to the clone builder to resolve @CaptaiNiveau's earlier comments

tanmay-hc (approved):

PR #302, opened 2024-11-21
WIP: [feat] add temporary SSH key generation

### LINKED ISSUE #301: Temporary SSH Key files not generated or used in proxmox-iso
#### Overview of the Issue

I am trying to avoid having to hardcode an SSH password or SSH key that will need to be changed/removed later.  In the documentation it says:

> If no communicator is defined, an SSH key is generated for use, and is used in the image's Cloud-Init settings for provisioning.

When not specifying any `communicator`, `ssh_username`, `ssh_password`, or `ssh_private_key_file`, it complains about needing an `ssh_username` (which matches the above statement about SSH being used by default; and the fact that the variable is required by the SSH communicator).

After supplying it with `ssh_username` I would expect it to generate a temporary key pair to be given to cloud-init, use those credentials to provision, and then remove them from the system.  However, it looks like it tries to authenticate with method "none" and inspecting the server created, the key is not in any `authorized_keys` files.

#### Reproduction Steps

In my specific instance I was creating an ubuntu-24.04.1 template from the official ISO and passing a CD with the `meta-data` and `user-data` (autoinstall) files (which should include cloud-init options).  I'm running Packer in the official Packer docker container with no modifications.

I run `init`, `validate`, and `build` in the docker container, and it works as expected until my above issue:
- download the image
- create the ISO containing `meta-data` and `user-data`
- upload both ISOs to Proxmox
- create a VM from the Ubuntu ISO with the cidata ISO mounted
- pass `boot_command` from source to instruct it to use the mounted cidata CD
- Ubuntu autoinstalls using the mounted CD
- VM reboots and is accessible by my host running Packer

#### Docker Version
```
v4.36.0
```

#### Packer Docker Image
```
light - f64016a1388f8d4b807978369bf09269b1823719562d4c78970a72ae10627b23
```

#### Packer Version
```bash
$ docker run --rm -e PACKER_LOG=1 hashicorp/packer:light version          
2024/11/19 02:44:15 [INFO] Packer version: 1.12.0-alpha1 [go1.21.12 linux amd64]
...
2024/11/19 02:44:15 machine readable: version []string{"1.12.0"}
2024/11/19 02:44:15 machine readable: version-prelease []string{"alpha1"}
2024/11/19 02:44:15 machine readable: version-commit []string{"9ff6fe7f"}
Packer v1.12.0-alpha1
...
```

#### proxmox Plugin Version
```
v1.2.1_x5.0_linux_amd64
```

#### Proxmox Version
```
pve-manager/8.2.8/a577cfa684c7476d
```

Removed everything that didn't seem pertinent to how keys should be created/deployed.  Can add more detail later if need be.

```hcl
packer {
    required_version = "~> 1"
    required_plugins {
        proxmox = {
            version = "~> 1"
            source = "github.com/hashicorp/proxmox"
        }
    }
}
source "proxmox-iso" "ubuntu-template" {
    qemu_agent = true
    additional_iso_files {
        cd_files = [
            "./files/autoinstall/meta-data",
            "./files/autoinstall/user-data"
        ]
        cd_label = "cidata"
        iso_storage_pool = "local"
    }
    cloud_init = true
    cloud_init_storage_pool = "local-lvm"
    boot_wait = "5s"
    boot_command = [
        "<esc><esc><esc><esc><wait>",
        "e<wait>",
        "<down><down><down><end>",
        " autoinstall ds=\"nocloud;\"<wait>",
        "<f10><wait>"
    ]
    # communicator = "ssh"
    ssh_username = "automation"
    # temporary_key_pair_type = "ed25519"
    ssh_timeout = "10m"
}

build {
    sources = ["source.proxmox-iso.ubuntu-template"]

    # Provisioning the VM Template for Cloud-Init Integration
    provisioner "shell" {
        inline = [
            "while [ ! -f /var/lib/cloud/instance/boot-finished ]; do echo 'Waiting for cloud-init...'; sleep 1; done",
            "sudo rm /etc/ssh/ssh_host_*",
            "sudo truncate -s 0 /etc/machine-id",
            "sudo apt -y autoremove --purge",
            "sudo apt -y clean",
            "sudo apt -y autoclean",
            "sudo cloud-init clean",
            "sudo rm -f /etc/cloud/cloud.cfg.d/subiquity-disable-cloudinit-networking.cfg",
            "sudo sync"
        ]
    }
    provisioner "file" {
        source = "files/99-pve.cfg"
        destination = "/tmp/99-pve.cfg"
    }
    provisioner "shell" {
        inline = [ "sudo cp /tmp/99-pve.cfg /etc/cloud/cloud.cfg.d/99-pve.cfg"]
    }
}
```

```
2024/11/19 01:50:11 [INFO] Packer version: 1.12.0-alpha1 [go1.21.12 linux amd64]
...
==> ubuntu-template.proxmox-iso.ubuntu-template: Waiting for SSH to become available...
2024/11/19 01:51:22 packer-plugin-proxmox_v1.2.1_x5.0_linux_amd64 plugin: 2024/11/19 01:51:22 [DEBUG] Unable to get address during connection step: 500 QEMU guest agent is not running   
2024/11/19 01:51:22 packer-plugin-proxmox_v1.2.1_x5.0_linux_amd64 plugin: 2024/11/19 01:51:22 [INFO] Waiting for SSH, up to timeout: 10m0s
2024/11/19 01:51:25 packer-plugin-proxmox_v1.2.1_x5.0_linux_amd64 plugin: 2024/11/19 01:51:25 [DEBUG] Error getting SSH address: 500 QEMU guest agent is not running
...
2024/11/19 01:58:14 packer-plugin-proxmox_v1.2.1_x5.0_linux_amd64 plugin: 2024/11/19 01:58:14 [DEBUG] Error getting SSH address: 500 QEMU guest agent is not running
2024/11/19 01:58:21 packer-plugin-proxmox_v1.2.1_x5.0_linux_amd64 plugin: 2024/11/19 01:58:21 [INFO] Attempting SSH connection to 192.168.209.233:22...
2024/11/19 01:58:21 packer-plugin-proxmox_v1.2.1_x5.0_linux_amd64 plugin: 2024/11/19 01:58:21 [DEBUG] reconnecting to TCP connection for SSH
2024/11/19 01:58:21 packer-plugin-proxmox_v1.2.1_x5.0_linux_amd64 plugin: 2024/11/19 01:58:21 [DEBUG] handshaking with SSH
2024/11/19 01:58:22 packer-plugin-proxmox_v1.2.1_x5.0_linux_amd64 plugin: 2024/11/19 01:58:22 [DEBUG] SSH handshake err: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none], no supported methods remain
2024/11/19 01:58:22 packer-plugin-proxmox_v1.2.1_x5.0_linux_amd64 plugin: 2024/11/19 01:58:22 [DEBUG] Detected authentication error. Increasing handshake attempts.
...
2024/11/19 02:01:15 packer-plugin-proxmox_v1.2.1_x5.0_linux_amd64 plugin: 2024/11/19 02:01:15 [INFO] Attempting SSH connection to 192.168.209.233:22...
2024/11/19 02:01:15 packer-plugin-proxmox_v1.2.1_x5.0_linux_amd64 plugin: 2024/11/19 02:01:15 [DEBUG] reconnecting to TCP connection for SSH
2024/11/19 02:01:15 packer-plugin-proxmox_v1.2.1_x5.0_linux_amd64 plugin: 2024/11/19 02:01:15 [DEBUG] handshaking with SSH
2024/11/19 02:01:15 packer-plugin-proxmox_v1.2.1_x5.0_linux_amd64 plugin: 2024/11/19 02:01:15 [DEBUG] SSH handshake err: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none], no supported methods remain
2024/11/19 02:01:15 packer-plugin-proxmox_v1.2.1_x5.0_linux_amd64 plugin: 2024/11/19 02:01:15 [DEBUG] Detected authentication error. Increasing handshake attempts.
==> ubuntu-template.proxmox-iso.ubuntu-template: Timeout waiting for SSH.
2024/11/19 02:01:22 packer-plugin-proxmox_v1.2.1_x5.0_linux_amd64 plugin: 2024/11/19 02:01:22 [DEBUG] SSH wait cancelled. Exiting loop.
...
2024/11/19 02:01:26 machine readable: error-count []string{"1"}
...
2024/11/19 02:01:26 machine readable: ubuntu-template.proxmox-iso.ubuntu-template,error []string{"Timeout waiting for SSH."}
...
Build 'ubuntu-template.proxmox-iso.ubuntu-template' errored after 11 minutes 12 seconds: Timeout waiting for SSH.
...
--> ubuntu-template.proxmox-iso.ubuntu-template: Timeout waiting for SSH.
...
```

bl1nk1n:
The documentation for the `proxmox-iso` plugin stated that when using the SSH communicator a temporary keypair will be generate if no other authorization methods are provided.  This operation did not take place and this commit aims to fix that using the SSH Communicator's 'StepSSHKeyGen' step.

Closes #301

```
**DELETE THIS TEMPLATE BEFORE SUBMITTING**

In order to have a good experience with our community, we recommend that you
read the contributing guidelines for making a PR, and understand the lifecycle
of a Packer Plugin PR:

https://github.com/hashicorp/packer-plugin-proxmox/blob/main/.github/CONTRIBUTING.md#opening-an-pull-request

Please include tests. We recommend looking at existing tests as an example. 
```

### DIFF
--- a/builder/proxmox/clone/builder.go
+++ b/builder/proxmox/clone/builder.go
@@ -39,10 +39,6 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 	state.Put("clone-config", &b.config)
 
 	preSteps := []multistep.Step{
-		&StepSshKeyPair{
-			Debug:        b.config.PackerDebug,
-			DebugKeyPath: fmt.Sprintf("%s.pem", b.config.PackerBuildName),
-		},
 		&StepMapSourceDisks{},
 	}
 	postSteps := []multistep.Step{}
--- a/builder/proxmox/common/builder.go
+++ b/builder/proxmox/common/builder.go
@@ -52,9 +52,11 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook,
 
 	// Build the steps
 	coreSteps := []multistep.Step{
+		// Since the ISOs get added to the VM before/during this line we need to update the cloud-init data before this line.  It probably needs to be before as CD creation is done in presteps
 		&stepStartVM{
 			vmCreator: b.vmCreator,
 		},
+		// Also need to update the cloud-init data before this line
 		commonsteps.HTTPServerFromHTTPConfig(&b.config.HTTPConfig),
 		&stepTypeBootCommand{
 			BootConfig: b.config.BootConfig,
@@ -67,14 +69,21 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook,
 		},
 		&commonsteps.StepProvision{},
 		&commonsteps.StepCleanupTempKeys{
-			Comm: &b.config.Comm,
+			Comm: comm,
 		},
 		&stepRemoveCloudInitDrive{},
 		&stepConvertToTemplate{},
 		&stepFinalizeTemplateConfig{},
 		&stepSuccess{},
 	}
-	preSteps := b.preSteps
+
+	preSteps := []multistep.Step{
+		&StepSshKeyPair{
+			Debug:		b.config.PackerDebug,
+			DebugKeyPath:	fmt.Sprintf("%s.pem", b.config.PackerBuildName),
+		},
+		&StepUpdateCloudInitSSH{},
+	}
 	for idx := range b.config.ISOs {
 		if b.config.ISOs[idx].ISODownloadPVE {
 			preSteps = append(preSteps,
@@ -104,8 +113,7 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook,
 		}
 	}
 
-	steps := append(preSteps, coreSteps...)
-	steps = append(steps, b.postSteps...)
+	steps := append(preSteps, b.preSteps..., coreSteps..., b.postSteps...)
 	// Run the steps
 	b.runner = commonsteps.NewRunner(steps, b.config.PackerConfig, ui)
 	b.runner.Run(ctx, state)
--- a/builder/proxmox/common/step_ssh_key_pair.go
+++ b/builder/proxmox/common/step_ssh_key_pair.go
@@ -1,7 +1,7 @@
 // Copyright (c) HashiCorp, Inc.
 // SPDX-License-Identifier: MPL-2.0
 
-package proxmoxclone
+package proxmox
 
 import (
 	"context"
@@ -47,9 +47,8 @@ func (s *StepSshKeyPair) Run(ctx context.Context, state multistep.StateBag) mult
 			return multistep.ActionHalt
 		}
 
-		c.Comm.SSHPrivateKey = privateKeyBytes
 		c.Comm.SSHKeyPairName = kp.Comment
-		c.Comm.SSHTemporaryKeyPairName = kp.Comment
+		c.Comm.SSHPrivateKey = privateKeyBytes
 		c.Comm.SSHPublicKey = kp.PublicKeyAuthorizedKeysLine
 
 		return multistep.ActionContinue
--- a/builder/proxmox/common/step_update_cloud_init_ssh.go
+++ b/builder/proxmox/common/step_update_cloud_init_ssh.go
@@ -1,0 +1,95 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package proxmox
+
+import (
+	"context"
+	"os"
+	"slices"
+	"strings"
+	"fmt"
+
+	packersdk "github.com/hashicorp/packer-plugin-sdk/packer"
+	common "github.com/hashicorp/packer-plugin-proxmox/builder/proxmox/common"
+	"github.com/hashicorp/packer-plugin-sdk/multistep"
+)
+
+type StepUpdateCloudInitSSH struct{}
+
+func (s *StepUpdateCloudInitSSH) Run(ctx context.Context, state multiste.StateBag) multistep.StepAction {
+	// NOTE: Can pass Cloud-Init data via CD or HTTP Server
+
+	ui := state.Get("ui").(packersdk.Ui)
+	c := state.Get("config").(*common.Config)
+
+	if c.Comm.SSHTemporaryKeyPairName == "" ||
+		(len(c.ISOs) <= 1 &&
+		 c.HTTPConfig.HTTPDir == "" &&
+		 len(c.HTTPConfig.HTTPContent) == 0) {
+		return multistep.ActionContinue
+	}
+
+	temp_ssh_public_key := string(c.Comm.SSHPublicKey)
+
+	if len(c.ISOs) > 1 {
+		// Skip first ISO as it should be the BootISO
+		for idx := range c.ISOs[1:] {
+			if c.ISOs[idx].CDConfig.CDLabel != "cidata" {
+				continue
+			} 
+			
+			if value, ok := c.ISOs[idx].CDConfig.CDContent["user-data"]; ok {
+				c.ISOs[idx].CDConfig.CDContent["user-data"] = strings.ReplaceAll(value, "${temporary_ssh_public_key}", temp_ssh_public_key)
+				ui.Say("Updated 'user-data' CD Content to use temporary SSH public key")
+				// CDContent will take precedence over CDFiles
+				break
+			}
+
+			// TODO: This needs to be able to handle when directories and globs 
+			// are passed to CDFiles
+			for jdx, value := range c.ISOs[idx].CDConfig.CDFiles {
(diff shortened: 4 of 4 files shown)


bl1nk1n:
I think after reading a TON more code I'm starting to understand something; but I have never written Go before, and I've only just touched Packer in the past week.  So please be generous with the constructive criticism! (Cross-posting on issue).

lbajolet-hashicorp:
Hey @bl1nk1n,

Thanks for the PR!

A few things come to mind:
1. Generating the SSH private key looks OK to me in the current state. I'm not certain type-checking for the communicator is the best way to do so, I see there's a step already defined in the `clone` builder, I wonder if we cannot reuse it? `builder/proxmox/clone/step_ssh_key_pair.go` is the one doing that.
2. Looking at the `clone` builder, it seems that the `Run` function for it is a good example of how to use that step, I would think we can replicate this on the `iso` builder as well, so long as the `StepSshKeyPair` is moved to `common`.
3. Just to be clear, the generation of the private key is one step, but there needs to be an extra step at some point to have the public key transmitted to the instance so the SSH connections with that key are accepted, I believe `clone` does that with `proxmoxapi.CloudInit`, unless I'm mistaken, you will probably need a similar mechanism for the `iso` builder.

Hope this feedback helps, let me know if something isn't clear!

bl1nk1n:
> [...] I'm not certain type-checking for the communicator is the best way to do so [...]

I just did this as the SSH keys should probably not be generated for other types of communicators such as `winrm`.  I based/copied it from the Packer communicator code that checks if it is `ssh` or `winrm` and calls the respective step.

> [...] I see there's a step already defined in the `clone` builder, I wonder if we cannot reuse it? [...]  Looking at the `clone` builder, it seems that the `Run` function for it is a good example of how to use that step, I would think we can replicate this on the `iso` builder as well, so long as the `StepSshKeyPair` is moved to `common`.

Ah!  I did not see that, I'll have to go take a look!  I haven't gotten to the step in my homelab IaC development to use `proxmox-clone` yet (literally my next step after building base templates with `proxmox-iso`; otherwise, I may have seen this.

> Just to be clear, the generation of the private key is one step, but there needs to be an extra step at some point to have the public key transmitted to the instance so the SSH connections with that key are accepted, I believe `clone` does that with `proxmoxapi.CloudInit`, unless I'm mistaken, you will probably need a similar mechanism for the `iso` builder.

I was wondering about this.  It felt super weird to just generate the SSH key and not add it in through CloudInit anywhere.  I was using other template preparation plugin code for other hypervisors as examples and didn't see where they added it into CloudInit.  So I assumed it was part of the process somehow (as I really don't know too much about Packer).

---

Thanks for the tips!  I'll try and work on this here in the following week or two.  While it is Thanksgiving, there is usually a decent amount of downtime with the family.

bl1nk1n:
@lbajolet-hashicorp 

I moved `StepSshKeyPair` to common and added another step to update Cloud-Init configurations that will be passed via HTTP or CD.  The code is rough and there are some parts where I am not sure what to do yet, but I wanted to give you an update in case there is feedback/direction you can give.  Again, sorry for the ignorance but this is my first time working with Go and Packer plugins.

The SSH key generation really was no problem, just like you stated.  However, what I noticed was that `proxmox-clone` injects the public key via Cloud-Init on VM creation; which won't necessarily work for `proxmox-iso`.  Usually Cloud-Init is not set up prior to template or even VM creation; it is done as a provisioning step.  Instead, the credentials would have to be passed in as part of something like Kickstart or Autoinstall via a configuration file.  These configuration files can be passed in via CD or the built-in HTTP server.  So for `proxmox-iso` the public key would more than likely need to be put into the configuration file.

Normally when using Packer I could just use `templatefile` and, given the SSH key as a variable, inject it into the configuration that way.  Instead, I needed to go through every way in which SSH keys can be passed in, pull the data down, replace the template variable `${temporary_ssh_public_key}` with the public key, and then write it either to memory or back to the file.  In the current state I have it, there are a few cases, when dealing with directories, in which this new feature won't work (yet).  If the user specifies a public key/password to be used and a temporary key is not generated, it will cause the template variable to not be filled (which is another issue).  In addition, it currently only looks for a `user-data` file/file contents which might only be for Ubuntu autoinstall.

In the current state, the user would basically specify `${temporary_ssh_public_key}` in their automated installation configuration and it will get filled in during build time.

I think this feature is possible to fully implement; however, after spending some time noodling with it, this might add unnecessary complexity and UI weirdness.  Trying to inject SSH keys into auto install configurations prior to CD creation/HTTP server deployment while supporting many OSes seems like it will get very messy.  This feature might need a more experienced touch.  If no one else is able to look at it, I am willing to continue learning Go/Packer plugins and messing with it with guidance!

bl1nk1n:
NOTE: I still haven't set up my development environment for this plugin/go/Packer so this code is not tested.  This will probably be my next priority now that I've laid a good foundation.

Issue #233, opened 2023-09-27
Re-open packer/issues/10227 to address issue with defaulting to IPv6


sier:
The issue with Packer attempting to use IPv6 is still an issue with the Proxmox provisioner (and I can only assume others as well).
The original issue has been closed without the second pull request to address this issue.
I wouldn't be opposed to having a switch option like "vm_interface_onlyipv4" for those masochists that have already adopted IPv6.

So far the only solution I can think of is to disable IPv6 completely, which isn't... ideal.

Original issue:

> The problem is 2-fold.
The used API client for Proxmox (Telmate/proxmox-api-go) cannot parse IPv6 addresses with a zone id (delimited by "%").
I've wrote a simple fix which splits the IPv6 address at the delimiter "%" and returns the first part.
That fix can be checked here: https://github.com/Telmate/proxmox-api-go/pull/112
> __Secondly I've added an extra check in the proxmox builder.go which basically tries to parse all IP addresses to IPv4 and thus returns the first valid IPv4 address.
I'll make the pull request.__

_Originally posted by @mmassez in https://github.com/hashicorp/packer/issues/10227#issuecomment-811838009_

PR #319, opened 2025-04-18
IPv6 support for the proxmox builder.

oromenahar:
Hey folks,

I added support for IPv6 and IPv4 dual stack setups or IPv6 only setups. The commit messages explains a lot for that reason I just copy the message to the description. I hope I didn't miss anything necessary and the PR can be discussed and reviewed.

It first checks for a valid IPv6 or IPv4 and continues if it's not valid. After that it continues if the ip is part of linkLocalUnicat, Multicast or loopback. The first ip which doesn't match one of the following types will be returned as valid.
no IPv4 or IPv6 at all
SiteLocalUnicast: fec0::/10
multicast: ff00::/8
LinkLocalUnicat: fe80::/64

If the proxmox interface supports IPv4 and IPv6 the selected ip depends on the proxmox list sorting of the returned addresses. The first valid ip will be used which can change depending on the proxmox api.

related to #233 
related to #10227
related to #10858

### DIFF
--- a/builder/proxmox/common/builder.go
+++ b/builder/proxmox/common/builder.go
@@ -176,10 +176,15 @@ func getVMIP(state multistep.StateBag) (string, error) {
 
 	for _, iface := range ifs {
 		for _, addr := range iface.IpAddresses {
-			if addr.IsLoopback() {
+			if addr.To4() == nil && addr.To16() == nil {
+				// no IPv4 or IPv6 at all
 				continue
-			}
-			if addr.To4() == nil {
+			} else if addr.IsLinkLocalUnicast() || addr.IsMulticast() || addr.IsLoopback() || addr.IsUnspecified() {
+				// can not be replaced by !addr.IsGlobalUnicast() because it just returns ips wich are globally routable,
+				// but Unique Local Addresses: fc00::/7 can be used as well for IPv6 setups.
+				// SiteLocalUnicast: fec0::/10
+				// multicast: ff00::/8
+				// LinkLocalUnicat: fe80::/64
 				continue
 			}
 			return addr.String(), nil


builder/proxmox/common/builder.go
```diff
@@ -176,10 +176,17 @@ func getVMIP(state multistep.StateBag) (string, error) {
...
+				// no IPv4 or IPv6 at all
 				continue
-			}
-			if addr.To4() == nil {
+			} else if addr.IsLinkLocalUnicast() || addr.IsMulticast() || addr.IsLoopback() {
+				// SiteLocalUnicast: fec0::/10
+				// multicast: ff00::/8
+				// multicast interface ff01::1
+				// multicast local: ff02::02
+				// Unique Local Unicast: fc00::/7
```
rgl: this one (ULU aka ULA) is not matched by `addr.IsLinkLocalUnicast() || addr.IsMulticast() || addr.IsLoopback()`, nor should it be. so please drop this comment.

builder/proxmox/common/builder.go
```diff
@@ -176,10 +176,17 @@ func getVMIP(state multistep.StateBag) (string, error) {
...
+				// global Unicast ::/96
```
rgl: this one is not matched by `addr.IsLinkLocalUnicast() || addr.IsMulticast() || addr.IsLoopback()`, nor should it be. so please drop this comment.

builder/proxmox/common/builder.go
```diff
@@ -176,10 +176,17 @@ func getVMIP(state multistep.StateBag) (string, error) {
 
 	for _, iface := range ifs {
 		for _, addr := range iface.IpAddresses {
-			if addr.IsLoopback() {
+			if addr.To4() == nil && addr.To16() == nil {
+				// no IPv4 or IPv6 at all
 				continue
-			}
-			if addr.To4() == nil {
+			} else if addr.IsLinkLocalUnicast() || addr.IsMulticast() || addr.IsLoopback() {
```
rgl: can't this be replaced with `!addr.IsGlobalUnicast()`?

also add `|| addr.IsUnspecified()`?

builder/proxmox/common/builder.go
```diff
@@ -176,10 +176,17 @@ func getVMIP(state multistep.StateBag) (string, error) {
...
+				// SiteLocalUnicast: fec0::/10
+				// multicast: ff00::/8
+				// multicast interface ff01::1
```
rgl: these two have no special meaning to the purpose of this code, and they are already covered by `multicast: ff00::/8`, so drop this line and the next?

builder/proxmox/common/builder.go
oromenahar: Added `|| addr.IsUnspecified()`

I don't think we should change this to `!addr.IsGlobalUnicast()` because this misses the ULA. We could change it to `!addr.IsGlobalUnicast() && !addr.IsPrivate()`? This would inlcude `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`, `fc00::/7` and all public IPs.

builder/proxmox/common/builder.go
```diff
@@ -176,10 +176,15 @@ func getVMIP(state multistep.StateBag) (string, error) {
...
+				// no IPv4 or IPv6 at all
 				continue
-			}
-			if addr.To4() == nil {
+			} else if addr.IsLinkLocalUnicast() || addr.IsMulticast() || addr.IsLoopback() || addr.IsUnspecified() {
+				// can not be replaced by !addr.IsGlobalUnicast() because it just returns ips wich are globally routable,
+				// but Unique Local Addresses: fc00::/7 can be used as well for IPv6 setups.
+				// SiteLocalUnicast: fec0::/10
+				// multicast: ff00::/8
+				// LinkLocalUnicat: fe80::/64
```
tanmay-hc, line 187: This changes the default communicator host selection from IPv4-only to the first usable IPv4 or IPv6 address. On dual-stack guests, that can silently switch existing builds to IPv6 if the guest agent reports it first. Can we preserve current behavior by preferring IPv4 and only falling back to IPv6 when no usable IPv4 exists?

builder/proxmox/common/builder.go
```diff
@@ -176,10 +176,15 @@ func getVMIP(state multistep.StateBag) (string, error) {
 
 	for _, iface := range ifs {
 		for _, addr := range iface.IpAddresses {
-			if addr.IsLoopback() {
+			if addr.To4() == nil && addr.To16() == nil {
+				// no IPv4 or IPv6 at all
 				continue
-			}
-			if addr.To4() == nil {
+			} else if addr.IsLinkLocalUnicast() || addr.IsMulticast() || addr.IsLoopback() || addr.IsUnspecified() {
```
tanmay-hc, line 182: Can we add focused tests for the address selection behavior here? We should cover IPv4-only, IPv6-only, and dual-stack ordering so this change is pinned down before merge.

Issue #328, opened 2025-07-10
proxmox-iso: Tag check is too strict, allowed character "." is rejected


dreibh:
I am trying to create a Proxmox VM with tag "ubuntu-24.04". This worked until recently, when I updated Packer and the plug-ins. Now, Packer refuses this Proxmox tag, due to "." in the tag.

#### Overview of the Issue

Build 'proxmox-iso' errored after 12 seconds 174 milliseconds: Error creating VM: tag may not start with - and may only include the following characters: abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_-

"ubuntu-24.04" is a valid tag, it should be accepted by Packer.

#### Reproduction Steps

  "builders": [
    {
      "type": "proxmox-iso",
      ...
      "tags": "ubuntu-24.04",

### Plugin and Packer version

dreibh@kringsjaa:~$ packer version
Packer v1.13.1
dreibh@kringsjaa:~$ packer plugin installed
/home/dreibh/.packer.d/plugins/github.com/hashicorp/docker/packer-plugin-docker_v1.1.1_x5.0_linux_amd64
/home/dreibh/.packer.d/plugins/github.com/hashicorp/proxmox/packer-plugin-proxmox_v1.2.2_x5.0_linux_amd64
/home/dreibh/.packer.d/plugins/github.com/hashicorp/qemu/packer-plugin-qemu_v1.1.3_x5.0_linux_amd64
/home/dreibh/.packer.d/plugins/github.com/hashicorp/virtualbox/packer-plugin-virtualbox_v1.1.2_x5.0_linux_amd64

### Log Fragments and crash.log files

2025-07-10T16:15:54+02:00: ==> proxmox-iso: Force set, checking for existing artifact on PVE cluster
2025/07/10 16:15:54 packer-plugin-proxmox_v1.2.2_x5.0_linux_amd64 plugin: 2025/07/10 16:15:54 vm '250710613' not found
2025-07-10T16:15:54+02:00: ==> proxmox-iso: No existing artifact found
2025-07-10T16:15:54+02:00: ==> proxmox-iso: Creating VM
2025-07-10T16:15:54+02:00: ==> proxmox-iso: Error creating VM: tag may not start with - and may only include the following characters: abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_-
[...]

Issue #339, opened 2025-11-11
Dependency update issue


dreibh:
The Proxmox plugin needs an update of  the https://github.com/Telmate/proxmox-api-go dependency to fix the tag regexp issue https://github.com/hashicorp/packer-plugin-proxmox/issues/328 . So, I tried updating the dependencies:
```
$ go get -u ./...
go: github.com/armon/go-metrics@v0.5.4: parsing go.mod:
        module declares its path as: github.com/hashicorp/go-metrics
                but was required as: github.com/armon/go-metrics
        trying github.com/armon/go-metrics@v0.5.3
go: github.com/armon/go-metrics@v0.5.3: parsing go.mod:
        module declares its path as: github.com/hashicorp/go-metrics
                but was required as: github.com/armon/go-metrics
        trying github.com/armon/go-metrics@v0.5.2
go: github.com/armon/go-metrics@v0.5.2: parsing go.mod:
        module declares its path as: github.com/hashicorp/go-metrics
                but was required as: github.com/armon/go-metrics
        trying github.com/armon/go-metrics@v0.5.1
go: github.com/armon/go-metrics@v0.5.1: parsing go.mod:
        module declares its path as: github.com/hashicorp/go-metrics
                but was required as: github.com/armon/go-metrics
        trying github.com/armon/go-metrics@v0.5.0
go: github.com/armon/go-metrics@v0.5.0: parsing go.mod:
        module declares its path as: github.com/hashicorp/go-metrics
                but was required as: github.com/armon/go-metrics
        trying github.com/armon/go-metrics@v0.4.2
go: github.com/armon/go-metrics@v0.4.2: parsing go.mod:
        module declares its path as: github.com/hashicorp/go-metrics
                but was required as: github.com/armon/go-metrics
        restoring github.com/armon/go-metrics@v0.4.1
```
Something seems to be seriously wrong with the dependencies. HashiCorp seems to have renamed dependency github.com/armon/go-metrics to github.com/hashicorp/go-metrics, but it is still required -- somewhere in the huge list of dependencies -- under its old name and version.

Building fails:
```
$ make
# github.com/hashicorp/packer-plugin-proxmox/builder/proxmox/common
[...]

PR #340, opened 2025-11-30
Update `Telmate/proxmox-api-go`

Tinyblargon:
### Description
What code changed, and why?

Hello, Maintainer of [Telmate/proxmox-api-go](https://github.com/Telmate/proxmox-api-go) here. Thought I'd help get this project over the context hurdle that was added in [976ef5098918469fca1cf843f96df391264410da](https://github.com/Telmate/proxmox-api-go/commit/976ef5098918469fca1cf843f96df391264410da).

Added the `context.Context` everywhere it was missing. Where I didn't know what the right call was I added `context.TODO()` directly.

Refactored the Qemu network interfaces to use the upstream [QemuNetworkInterfaces](https://github.com/Telmate/proxmox-api-go/blob/81c14ce502d86ccf7e905ee35e9a705eb2874d35/proxmox/config__qemu__network.go#L354).

Refactored the Qemu PCI devices to use the upstream [QemuPciDevices](https://github.com/Telmate/proxmox-api-go/blob/81c14ce502d86ccf7e905ee35e9a705eb2874d35/proxmox/config__qemu__pcie.go#L11). However there is a problem there. The following properties have not yet been implemented upstream:

- `mdev`
- `romfile`
- `legacy-igd`

I'm of the understanding these features are rarely used. As someone would have opened a ticket for it otherwise.

For `mdev` we have an [issue](https://github.com/Telmate/proxmox-api-go/issues/526) and it will be fixed.
However for `romfile` and `legacy-igd` we are unsure how they would fit in our current data structure.

### Resolved Issues
If your PR resolves any open issue(s), please indicate them like this so they will be closed when your PR is merged:

Related to #339

### Rollback Plan

n/a

### Changes to Security Controls

n/a

### DIFF
--- a/builder/proxmox/clone/builder.go
+++ b/builder/proxmox/clone/builder.go
@@ -53,7 +53,7 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 
 type cloneVMCreator struct{}
 
-func (*cloneVMCreator) Create(vmRef *proxmoxapi.VmRef, config proxmoxapi.ConfigQemu, state multistep.StateBag) error {
+func (*cloneVMCreator) Create(ctx context.Context, vmRef *proxmoxapi.VmRef, config proxmoxapi.ConfigQemu, state multistep.StateBag) error {
 	client := state.Get("proxmoxClient").(*proxmoxapi.Client)
 	c := state.Get("clone-config").(*Config)
 	comm := state.Get("config").(*proxmox.Config).Comm
@@ -140,7 +140,7 @@ func (*cloneVMCreator) Create(vmRef *proxmoxapi.VmRef, config proxmoxapi.ConfigQ
 
 	var sourceVmr *proxmoxapi.VmRef
 	if c.CloneVM != "" {
-		sourceVmrs, err := client.GetVmRefsByName(c.CloneVM)
+		sourceVmrs, err := client.GetVmRefsByName(ctx, c.CloneVM)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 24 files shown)


3deep5me:
would be nice to get this merged to fix the issue #328 - can i help somehow?

rgl:
Ping?

rgl:
Ping?

rnsc:
Hello @tanmay-hc @domrim @mpywell @hariom-hashicorp 

I'm sorry for the direct pings, but due to this plugin not keeping up with the proxmox-api-go it's using, when building templates in Proxmox, we are faced with various issues that have been fixed upstream.

One example is that we cannot use custom cpu types when building VMs/templates while it's supported by proxmox-api-go ie: https://github.com/Telmate/proxmox-api-go/commit/13c81efcbbbe59c2e56794389901195dcd57b37c

I understand that maybe there are different priorities at Hashicorp, but it'd be nice if the community contributions could move forward.

Thank you,

tanmay-hc:
Hey @rnsc,
We fell behind on maintaining some plugins for a whle, but we're actively working through updates lately.
Right now, the PR is waiting on @Tinyblargon to address the Copilot feedback and rebase before we can proceed.
Our team has limited domain expertise with Proxmox, so we rely heavily on the community and tooling like Copilot to help fill knowledge gaps and flag issues during review.

Tinyblargon:
@tanmay-hc I'll address the feedback from copilot and rebase.

There is one change related to pcie where packer exposes more settings than the proxmox library. How should I continue with that? Silenty drop, wait till they are implented upstream, drop and log a warning? Not sure the last one is possible.

Also should I try to update the library as far as possible? Or get it working till this point and make a new PR for further updates?

tanmay-hc:
@Tinyblargon Thanks for flagging the PCIe gap instead of quietly working around it.

I went through the code and one thing to note: splitting the PCI commit out isn't an option. The bump to `976ef509` drops `ConfigQemu.QemuPCIDevices`, which is what made `mdev`, `romfile` and `legacy_igd` work in the first place. Since that's the same commit that adds the context changes, we can't take one without the other. Dropping your PCI commit doesn't build:
```
step_start_vm.go:150:3: unknown field QemuPCIDevices in struct literal of type ConfigQemu
```
On the three fields: silently dropping is out. They've been documented and released since v1.1.4, so existing templates would keep building but produce a misconfigured VM. `legacy_igd` is the worst of them, `Prepare` still validates it against machine type and VGA, so we'd tell the user it's supported and then ignore it.

Warnings are possible as we use them for deprecations, but the build still succeeds with a broken image, so I don't think that's enough for passthrough.

So I'd rather wait for upstream. Since you maintain the library, would adding these to `QemuPciRaw`/`QemuPciMapping` be workable? If `romfile` and `legacy-igd` are awkward to model, even a field for unmodelled keys would let this land without a regression.

If waiting isn't practical, the fallback is a hard error in `Prepare` when any of the three is set, pointing at the tracking issue. Still breaking, so it'd need a CHANGELOG note.

Let me know your thoughts

PR #351, opened 2026-04-05
[Issue 346] Respect HTTPS_PROXY and HTTP_PROXY environment variables

### LINKED ISSUE #346: Proxmox API client ignores HTTPS_PROXY environment variable
## Summary

The Proxmox packer plugin does not respect `HTTPS_PROXY` / `HTTP_PROXY` environment variables when connecting to the Proxmox API. This makes it impossible to use the plugin in environments where the Proxmox server is only reachable through an HTTP proxy (e.g., on-prem Proxmox behind a VPN proxy from a cloud-hosted CI runner).

## Root Cause

In `builder/proxmox/common/client.go`, `newProxmoxClient()` calls `proxmox.NewClient()` with an empty proxy string:

```go
client, err := proxmox.NewClient(strings.TrimSuffix(config.proxmoxURL.String(), "/"), nil, "", tlsConfig, "", int(config.TaskTimeout.Seconds()))
```

The upstream Telmate library (`proxmox-api-go`) then creates an `http.Transport` with `Proxy: nil` when the proxy string is empty:

```go
// proxmox/session.go in NewSession()
if proxyString == "" {
    tr := &http.Transport{
        TLSClientConfig:    tls,
        DisableCompression: true,
        Proxy:              nil,  // explicitly disables proxy
    }
```

Setting `Proxy: nil` explicitly disables Go's default proxy behavior (`http.ProxyFromEnvironment`), which means `HTTPS_PROXY`, `HTTP_PROXY`, and `NO_PROXY` environment variables are completely ignored.

## Expected Behavior

When no explicit proxy is configured, the plugin should respect standard proxy environment variables (`HTTPS_PROXY`, `HTTP_PROXY`, `NO_PROXY`) by using `http.ProxyFromEnvironment` as the proxy function on the HTTP transport.

## Proposed Fix

Two options (either would work):

**Option A** — Fix in packer-plugin-proxmox: create an `http.Client` with `http.ProxyFromEnvironment` and pass it to `proxmox.NewClient()` instead of `nil`:

```go
tr := &http.Transport{
    TLSClientConfig: tlsConfig,
    Proxy:           http.ProxyFromEnvironment,
}
hclient := &http.Client{Transport: tr}
client, err := proxmox.NewClient(..., hclient, "", tlsConfig, "", ...)
```

**Option B** — Fix in Telmate/proxmox-api-go: change the default `Proxy` from `nil` to `http.ProxyFromEnvironment` when no proxy string is provided.

## Environment

- packer-plugin-proxmox v1.2.3
- Running in Kubernetes (EKS) with an HTTP proxy for VPN access to on-prem Proxmox
- `HTTPS_PROXY` is set and works correctly with curl and other Go HTTP clients

lloydmcl:
### Description

The plugin currently passes `nil` as the HTTP client to `proxmox.NewClient()`.
The upstream library's `NewSession()` handles this by creating a transport with
`Proxy: nil`, which explicitly disables Go's standard proxy environment variable
handling. Users behind corporate proxies cannot use the plugin even when
`HTTPS_PROXY` is correctly configured.

The root cause is in `session.go` of `proxmox-api-go` — when `hclient` is `nil`
and `proxyString` is empty, it builds a transport with `Proxy: nil` rather than
`http.ProxyFromEnvironment`.

This change creates a configured `http.Client` with `http.ProxyFromEnvironment`
and passes it to `proxmox.NewClient()`. Since a non-nil client is provided, the
upstream library uses it directly, bypassing its own transport creation. When no
proxy environment variables are set, behaviour is identical to the current
implementation.

This follows the same approach used by Terraform providers and other Go HTTP
clients for proxy support.

### Resolved Issues

Closes #346

### Rollback Plan

If a change needs to be reverted, we will roll out an update to the code within 7 days.

### Changes to Security Controls

HTTP proxy configuration is read from standard environment variables
(`HTTPS_PROXY`, `HTTP_PROXY`, `NO_PROXY`). This is standard Go behaviour via
`http.ProxyFromEnvironment` and does not introduce new attack surface — the
proxy must be explicitly configured in the user's environment.

### DIFF
--- a/builder/proxmox/common/client.go
+++ b/builder/proxmox/common/client.go
@@ -6,6 +6,7 @@ package proxmox
 import (
 	"crypto/tls"
 	"log"
+	"net/http"
 	"strings"
 
 	"github.com/Telmate/proxmox-api-go/proxmox"
@@ -16,7 +17,14 @@ func newProxmoxClient(config Config) (*proxmox.Client, error) {
 		InsecureSkipVerify: config.SkipCertValidation,
 	}
 
-	client, err := proxmox.NewClient(strings.TrimSuffix(config.proxmoxURL.String(), "/"), nil, "", tlsConfig, "", int(config.TaskTimeout.Seconds()))
+	httpClient := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: tlsConfig,
+			Proxy:           http.ProxyFromEnvironment,
+		},
+	}
+
+	client, err := proxmox.NewClient(strings.TrimSuffix(config.proxmoxURL.String(), "/"), httpClient, "", tlsConfig, "", int(config.TaskTimeout.Seconds()))
 	if err != nil {
 		return nil, err
 	}


lloydmcl:
Apologies — I missed PR #347 which addresses the same issue. Closing in favour of that one. Cheers.

